PR #100, opened 2025-09-12
feat: use blockwise scaling for V in FP8 FA triton kernel

hann-wang:
As discussed with Carlus, the same quantization scheme for Q, K, V is preferred. This PR enables blockwise scaling for V.


kyle-256 (approved):

hann-wang:
MLPerf Llama2-70B LoRA finetuning loss:

image

xiaobochen-amd (approved):

PR #108, opened 2025-09-22
[Fix] Building on CPU nodes should not infer device arch

GeneDer:
Building on CPU nodes should not infer device arch. Lazy import torch and infer the device during setup only if `GPU_ARCHS` is not set.


xiaobochen-amd:
LGTM

PR #114, opened 2025-09-24
feat(token_dispatcher): add TurboDeepEPTokenDispatcher/Combiner and TurboTokenPermuter/Unpermuter of MoE.

zhenhuang12:
## TurboDeepEPTokenDispatcher/Combiner

The `TurboDeepEPTokenDispatcher/Combiner` torch modules provide MoE dispatch/combine modules for user.  Megatron's `MoETokenDispatcher` has a nice API, but it's not a torch module. We can inherits MoETokenDispatcher and torch module, but it's weird when we use `MoETokenDispatcher` api instead of torch module's `forward`.

The capability of both dispatcher and combiner are similar to Megatron's `MoETokenDispatcher` . The dispatcher is equivalent to dispatch + permute, and the combiner is equivalent to unpermute + combine.

The main differences are:
- dispatcher/combiner are parameterless torch modules, making them simpler to use with other modules.
- support for Sync-Free MoE, allowing the dispatcher to avoid host sync via options.

If we want to implement shared experts overlap just like Megatron's `MoEAlltoAllTokenDispatcher`, we can add forward and backward hooks to dispatcher/combine and permuter/unpermute to achieve overlap, or just use turbo's ops.

### DIFF
--- a/primus_turbo/pytorch/modules/moe/__init__.py
+++ b/primus_turbo/pytorch/modules/moe/__init__.py
@@ -1,0 +1,6 @@
+from .token_dispatcher import (
+    TurboDeepEPTokenCombiner,
+    TurboDeepEPTokenDispatcher,
+    TurboTokenPermuter,
+    TurboTokenUnpermuter,
+)
--- a/tests/pytorch/modules/test_token_dispatcher.py
+++ b/tests/pytorch/modules/test_token_dispatcher.py
@@ -5,6 +5,8 @@
 
 
 from dataclasses import dataclass
+from functools import lru_cache
+from itertools import product
 
 import torch
 import torch.distributed as dist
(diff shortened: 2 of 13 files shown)


primus_turbo/pytorch/modules/moe/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .token_dispatcher import (
+    TurboDeepEPTokenCombiner,
```
RuibinCheung, line 2: Can you simplify the class name from `TurboDeepEPTokenCombiner` to `DeepEPTokenCombiner` ? 
I think the user can identify it from TE or Primus-Turbo when importing (e.g `import primus_turbo.xxx.DeepEPTokenDispatcher as TurboDeepEPTokenDispatcher`).

tests/pytorch/modules/test_token_dispatcher.py
```diff
@@ -0,0 +1,169 @@
...
+import primus_turbo.pytorch as turbo
+
+
+@dataclass
+class TokenDispatcherTestConfig:
+    # random data generation
+    num_tokens: int
+    hidden_size: int
+    dtype: torch.dtype
+    # TurboDeepEPTokenDispatcher init
```
RuibinCheung, line 28:
```suggestion
    # TokenDispatcher init
```

tests/pytorch/modules/test_token_dispatcher.py
```diff
@@ -0,0 +1,169 @@
...
+    router_topk: int
+    num_experts: int
+    permute_fusion: bool
+    deepep_use_cuda_num_tokens_per_expert: bool
+
+    # TurboDeepEPTokenDispatcher forward
```
RuibinCheung, line 34: Same as above.

PR #119, opened 2025-09-26
remove -fgpu-rdc and --hip-link args

wenxie-amd:


xiaobochen-amd (approved):

zhenhuang12:
why remove `fgpu-rdc` and `hip-link`? these flags are also needed by DeepEP with rocSHEMM.

PR #121, opened 2025-09-28
feat: gemm fp8 support cktile backend for both tensorwise and rowwise

kyle-256:


primus_turbo/pytorch/core/float8.py
```diff
@@ -98,6 +98,7 @@ class Float8QuantConfig:
     granularity: ScalingGranularity = ScalingGranularity.TENSORWISE
     strategy: ScalingStrategy = ScalingStrategy.DYNAMIC
     block_size: Optional[int] = None  # Default: not used for tensorwise/rowwise
+    backend: str = "ck"  # "hipblaslt" or "ck"
```
xiaobochen-amd, line 101: It might be better if we avoid exposing the backend to users.

PR #138, opened 2025-10-22
feat(deepep): move deep_ep header file to primus_turbo common header dir

zhenhuang12:


zhenhuang12:
> If api.cuh only contains host logic and no device code (like **global**, **device**), then it’s better to use api.h instead? The .cuh, I think it is mainly for files that include or declare GPU-side code.

great! I change .h to .cuh in order to align with DeepEP. I will change it's suffix back.

xiaobochen-amd (approved):

PR #156, opened 2025-11-14
feat: add float4x2_e2m1 and float8_m8m0 data type

xiaobochen-amd:


kyle-256 (dismissed):

primus_turbo/pytorch/core/low_precision.py
```diff
@@ -72,11 +89,13 @@ def check_fp8_ocp_support() -> Tuple[bool, str]:
 
 class Format(Enum):
     """
-    Supported FP8 formats.
+    Supported FP8/FP4 formats.
     """
 
     E4M3 = auto()
     E5M2 = auto()
+    E8M0 = auto()
```
RuibinCheung: What's the meaning of E8M0 format ?

RuibinCheung (changes requested):

primus_turbo/pytorch/core/low_precision.py
xiaobochen-amd: Are there any cases where we need `Format` to specify e8m0? I can’t think of any at the moment. If not, we can remove it.

RuibinCheung (approved):
LGTM

PR #162, opened 2025-11-24
feat(deepep): support single-node moe_dispatch/combine on jax.

zhenhuang12:


csrc/jax/deep_ep/deep_ep.cpp
yeandy: Is most of the logic here matching the code from Pytorch's cpp DeepEP implementation https://github.com/AMD-AGI/Primus-Turbo/blob/main/csrc/pytorch/deep_ep/deep_ep.cpp, but just adapted for JAX? 

As an aside, I noticed variables here ending with `_`. i.e. `barrier_signal_ptrs_gpu_` here, but not in the pytorch implementation. Can you help me understand the different usage in syntax here?

tests/jax/lax/test_dispatch_combine.py
```diff
@@ -0,0 +1,263 @@
...
+    topk_weights = jnp.ones((num_ranks, num_tokens, num_topk), dtype=jnp.float32)
+
+    for i in range(num_ranks):
+        x = x.at[i].set(i)
+        topk_weights = topk_weights.at[i].set(i)
+    return (
+        jnp.reshape(x, (num_ranks * num_tokens, hidden)),
+        jnp.reshape(scores, (num_ranks * num_tokens, num_experts)),
+        jnp.reshape(topk_weights, (num_ranks * num_tokens, num_topk)),
+    )
```
yeandy: Can we add a comment on how the different items in the tuples correspond to B, M, N, K?

primus_turbo/jax/lax/moe/moe_dispatch_combine.py
```diff
@@ -0,0 +1,290 @@
...
+        num_ranks: the number of ranks.
+
+    Returns:
+        config: the recommended config.
+    """
+    global _default_num_sms
+    return Config(_default_num_sms, 36, 288, 20, 128)
+
+
+def get_combine_config(num_ranks: int = 8) -> Config:
```
yeandy: how is `num_ranks` used inside these two functions?

csrc/jax/deep_ep/deep_ep.cpp
zhenhuang12: > Is most of the logic here matching the code from Pytorch's cpp DeepEP implementation https://github.com/AMD-AGI/Primus-Turbo/blob/main/csrc/pytorch/deep_ep/deep_ep.cpp, but just adapted for JAX?
> 
> As an aside, I noticed variables here ending with `_`. i.e. `barrier_signal_ptrs_gpu_` here, but not in the pytorch implementation. Can you help me understand the different usage in syntax here?

yeah, most of logic is the same as torch's DeepEP, but some functions are difference on jax.

I followed Google style when writing Jax's DeepEP, but PyTorch still follows the default DeepEP official specification. Do you have any suggestions about code style?

primus_turbo/jax/lax/moe/moe_dispatch_combine.py
zhenhuang12: `get_dispatch_config` and ` get_combine_config` align with torch, but currently only supports num_ranks <= 8.

primus_turbo/jax/lax/moe/moe_dispatch_combine.py
```diff
@@ -0,0 +1,290 @@
...
+    moe_dispatch_p,
+)
+
+from .moe_utils import Config
+
+__all__ = ["get_dispatch_config", "moe_dispatch", "get_combine_config", "moe_combine"]
+
+
+_default_num_sms = 64
+num_ranks = 8
```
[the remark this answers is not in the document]
zhenhuang12: I used `jax.local_device_count()` replace with hard code `num_ranks`, only support single-node for now. We should pass `jax.lax.axis_index` (ep rank) and `jax.lax.axis_size` (ep world size) to the C++ ffi_call, but it change to `jax._src.interpreters.partial_eval.DynamicJaxprTracer` after being processed by `jax.jit`, which not support for ffi_call

PR #208, opened 2026-01-14
feat(attn): support sink attention

kyle-256:


xiaobochen-amd:
**Design Proposal**

Treat the AITER C++ implementation and the Triton implementation as two schedulable backends.
In attention_aiter_fwd/bwd_impl, dispatch through a unified AttnFwd/BwdAiterKernelDispatch.

**Dispatch policy (auto-tune can be deferred for now):**
* Prefer the C++ backend
* If the C++ backend cannot support the current configs, fallback to the Triton backend
* If the Triton backend also cannot handle it, raise an error

**Benefits:**
* Clearly separates responsibilities between different implementations
* Avoids branching and duplicated logic at the Python layer
* Enables simple and effective reuse of aiter-triton in CP scenarios

setup.py
```diff
@@ -383,13 +383,13 @@ def build_jax_extension():
     ]
 
     # Conditionally add aiter if torch_ext is being built and aiter is not already installed
-    if torch_ext is not None and not is_package_installed("aiter"):
-        print("[Primus-Turbo Setup] aiter not found, will be installed automatically.")
+    if torch_ext is not None and not is_package_installed("amd-aiter"):
+        print("[Primus-Turbo Setup] amd-aiter not found, will be installed automatically.")
         install_requires.append(
-            "aiter @ git+https://github.com/ROCm/aiter.git@a25cb58ee4ee97cb5ad4b426b648a8e66103d706"
+            "amd-aiter @ git+https://github.com/ROCm/aiter.git@311deb04d918d10449d57f780295ee487dff5521"
```
xiaobochen-amd: Has aiter been renamed to amd-aiter?

kyle-256: Yes

primus_turbo/triton/moe/permutation.py
```diff
@@ -10,13 +10,35 @@
...
 import triton.language as tl
-from triton.language import core
 from triton.language.standard import _log2
 
 # The following three argsort related kernels are adapted from
 # the issue https://github.com/triton-lang/triton/issues/3698
 
 
+@triton.jit
+def _to_int_bitcast(val, signed: tl.constexpr = True):
```
xiaobochen-amd: The same code is written in two places, which is somewhat redundant.

xiaobochen-amd (approved):

PR #209, opened 2026-01-15
[Aiter] upgrade aiter to be compatible with latest TE

GeneDer:
TE recently updated the aiter commit they are using https://github.com/ROCm/TransformerEngine/pull/354 ([aiter @ ed4bdea](https://github.com/ROCm/aiter/tree/ed4bdeac27b930f8efb1976af14db88ba3d648ad)). There is an api change around calling `_flash_attn_forward`. This PR tries to use the new api offered in aiter.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -382,15 +382,14 @@ if __name__ == "__main__":
         "hip-python",
     ]
 
-    # Conditionally add aiter if torch_ext is being built and aiter is not already installed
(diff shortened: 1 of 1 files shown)


xiaobochen-amd:
@GeneDer  FYI: this PR also needs to update the AITER version.

https://github.com/AMD-AGI/Primus-Turbo/pull/208

PR #236, opened 2026-02-11
fix: force to use fgpu-rdc to compile

zhenhuang12:
From JIRA [AIMA-212](https://amd.atlassian.net/browse/AIMA-212), it shows that DeepEP intranode-combine timeout on rocm7.2 docker image, but works well on rocm7.1.

I found a workaround to solve the timeout which forcely add `-fgpu-rdc` to compile and disable `-amdgpu-function-calls=false`


xiaobochen-amd:
I remember this compile option caused issues before.

xiaobochen-amd (approved):

PR #249, opened 2026-03-16
feat(mxfp4): integrate aiter fp4 gemm and support hip mxfp4 quant and shuffle kernel

RuibinCheung:
Function:
* Integrate aiter fp4 gemm into turbo
Kernels:
* Add hip mxfp4 quant kernel
* Add hip shuffle kernel
Core: 
* Enhance `GlobalBackendManager's` function
* Add logger module


primus_turbo/common/logger.py
xiaobochen-amd: The current logger is implemented as multiple independent functions (e.g., log_debug, log_info_rank0, log_debug_once), which results in a relatively large number of functions and combinations. This can make maintenance slightly cumbersome.

It could be refactored into a singleton class, with a design along the following lines:
```
import logging

class PrimusTurboLogger:
    def __init__(self, name="primus_turbo"):
        ...
    
    def log(self, level, msg, once=False, rank=None):
        ...

    def info(self, msg, once=False, rank=None):
        return self.log(logging.INFO, msg, once, rank)

    def debug(self, msg, once=False, rank=None):
        return self.log(logging.DEBUG, msg, once, rank)

    ......


# Global instance
logger = PrimusTurboLogger(...)
```

Usage would then be as simple as importing this instance:
```
from primus_turbo.common import logger

logger.info("xxxx")
logger.info("yyyy", once=True)
```

xiaobochen-amd (approved):

PR #275, opened 2026-04-08
feat: attn support sbhd layout

RuibinCheung:
Attention support sbhd layout

NOTE: 
1. Sink attn kernel not support sbhd layout. Need to fallback to BSHD layout on Primus side.


primus_turbo/pytorch/ops/attention/flash_attn_interface.py
```diff
@@ -54,17 +48,17 @@ def forward(
         return_lse,
         return_softmax,
         is_grad_enabled,
-        is_v3_atomic_fp32: Optional[bool] = None,
         how_v3_bf16_cvt: Optional[int] = 1,
         sink: Optional[torch.Tensor] = None,
+        qkv_format: Optional[str] = "bshd",
```
xiaobochen-amd, line 54: We could add a helper function to check whether the input layout matches the expected format based on qkv_format. This would improve robustness. If the input does not match qkv_format, then forcibly convert it to qkv_format.

RuibinCheung:
@cursor review

RuibinCheung:
@cursor review

primus_turbo/pytorch/ops/attention/flash_attn_interface.py
RuibinCheung: ACK. Thx.

xiaobochen-amd (approved):

PR #277, opened 2026-04-10
feat: Online tuning for hipblaslt gemm

Z-Y00:


docs/examples.md
```diff
@@ -490,16 +490,31 @@ GlobalBackendManager.set_gemm_backend(None)
...
+NOTE: Setting an explicit backend (via `set_gemm_backend()` or `PRIMUS_TURBO_GEMM_BACKEND`) takes priority over AutoTune at any level. If you want level-2 algo tuning, do **not** set an explicit backend — let AutoTune select the backend and algorithm together.
+
 ```python
 import torch
 import primus_turbo.pytorch as turbo
 from primus_turbo.pytorch.core.backend import GlobalBackendManager
 
-GlobalBackendManager.set_auto_tune(True)  # or set PRIMUS_TURBO_AUTO_TUNE=1
+# Level 1: backend selection only (same as the old True / PRIMUS_TURBO_AUTO_TUNE=1)
+GlobalBackendManager.set_auto_tune(1)
```
RuibinCheung, line 511: I think the reuse of `set_auto_tune` will cause confused. I suggest that you can add another an API named `GlobalBackendManager.set_tune_level()` to control tuning level.
1. When `set_auto_tune(False)`, disable auto tune.
2. When `set_auto_tune(True)` and `set_tune_level(1)`, use backend selection only
3. When `set_auto_tune(True)` and `set_tune_level(2)`, use backend selection and hipblaslt multi-algo tuning.

docs/examples.md
```diff
@@ -490,16 +490,31 @@ GlobalBackendManager.set_gemm_backend(None)
...
 
+NOTE: Setting an explicit backend (via `set_gemm_backend()` or `PRIMUS_TURBO_GEMM_BACKEND`) takes priority over AutoTune at any level. If you want level-2 algo tuning, do **not** set an explicit backend — let AutoTune select the backend and algorithm together.
+
 ```python
 import torch
 import primus_turbo.pytorch as turbo
 from primus_turbo.pytorch.core.backend import GlobalBackendManager
 
-GlobalBackendManager.set_auto_tune(True)  # or set PRIMUS_TURBO_AUTO_TUNE=1
+# Level 1: backend selection only (same as the old True / PRIMUS_TURBO_AUTO_TUNE=1)
```
RuibinCheung, line 510: Also add an environmental variable to control auto tune level.

RuibinCheung (changes requested):

xiaobochen-amd:
The new feature added in primus_turbo/pytorch/core/backend.py::AutoKernelDispatcher is too biased toward hipblaslt. It’s not recommended to add it here. With a different backend, it may no longer be an algo_index.

benchmark/ops/bench_hipblaslt_algo_tuning.py
xiaobochen-amd, line 1: This benchmark file can be removed. We should try to consolidate everything into the existing bench_gemm_turbo.py as much as possible.

xiaobochen-amd:
Can we remove hipblaslt_gemm_algo_count?

We could replace it with something like:
online_tune_hipblaslt_gemm(..., max_num_tune=50), which directly returns the best algo_index (feel free to decide the exact function name).

This would mean that when Level 2 is enabled, it triggers online_tune_hipblaslt_gemm, obtains the best result, and writes it into the cache.

The idea is that the upper-level framework doesn’t need to be aware of how many algos each backend has. It only needs to specify a maximum tuning budget and retrieve the best-performing result for the current backend.

PR #300, opened 2026-04-21
fix(grouped_gemm): stabilize hipBLASLt multi-stream for BF16 and FP8

kyle-256:
# Multi-Stream hipBLASLt Grouped GEMM Stalls

Two independent root causes for intermittent stalls of `HipblasltGroupedGemm`
under `kMaxNumStreams=4` multi-stream execution. Both fixes preserve full
4-stream parallelism.

## BF16 — fixed in `d50df95`

**Root cause**: Internal `hipblasLtHandle_t`s for slots 1..3 were never bound
to their compute streams via `hipblasSetStream`. hipBLASLt's per-handle
internal state (heuristic cache, workspace accounting) assumes a 1:1
handle↔stream relationship. Reusing one handle from multiple streams races
on this state, intermittently serializing kernels — DeepSeek-V3-Down
backward jumped from ~4 ms to 300-1400 ms per iter.

**Fix** (`csrc/kernels/grouped_gemm/hipblaslt_grouped_gemm.cu`):

- `hipblasSetStream(handles_[i], compute_streams_[i])` at construction for
  each internal handle (1:1 binding).
- Slot 0 keeps `params.handle` + `params.stream` (PyTorch's, already 1:1
  bound).
- Skip the cross-stream `sync_event_` fan-out/fan-in when
  `num_stream_used == 1`; loop starts at `s=1` so slot 0 isn't
  double-synchronized.

Matches the ROCm TransformerEngine and grouped-gemm-ck reference
implementations.

## FP8 — fixed in `7743364`

**Root cause** (found via rocprofv3 trace comparison): the FP8 Stream-K
persistent kernel
`Cijk_..._F8BS_..._MT256x256x128_MI16x16x1_CMS_SN` (note `SK3` / `_SN`
suffixes) deadlocks when 4 concurrent instances run on different streams —
same kernel binary, same WG/VGPR/SGPR/LDS, but single-wg execution time
jumps from ~495 µs to 25 s, and all 4 stream instances stall and recover
together. The cross-CU atomic coordination Stream-K uses to partition
K-axis work has a bad concurrency pattern when multiple instances overlap.

The grid size is heuristic-driven by
`HIPBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES`:

| Workspace hint | Grid_Size_X | per-wg time | 4-stream concurrent |
|----------------|-------------|-------------|---------------------|
| 64 MiB         | 62 720      | ~495 µs     | **deadlocks**       |
| 8 MiB          | 94 208      | ~125 µs     | **stable**          |

Smaller K-axis chunks → shorter atomic-coordination windows → no deadlock.

**Fix** (`csrc/kernels/gemm/hipblaslt_gemm.cu`, one line):

```cpp
int64_t pref_workspace_size = use_low_precision ? int64_t{8 * 1024 * 1024}
                                                : workspace_size;
```

Only the heuristic *hint* is reduced. The actual workspace passed to
`hipblasLtMatmul` is unchanged. BF16 keeps the full 64 MiB hint and is
unaffected (its kernels aren't Stream-K persistent on these shapes).

## Verification (MI355X, gfx950)

| Test                                                          | Result                              |
|---------------------------------------------------------------|-------------------------------------|
| 80 000 fwd+bwd measurements (16 shapes × 100 iter × 5 passes) | 0 stalls                            |
| 1 200 measurements of previous worst case (was 25 s stalls)   | 3.73-4.17 ms, no spikes             |
| 10 official FP8 benchmark runs                                | avg fwd 1084-1102 TFLOPS, 1.7% spread |
| FP8 vs prior baseline                                         | +13% fwd, +5% bwd                   |
| BF16 benchmark                                                | unchanged                           |
| `kMaxNumStreams`                                              | stays 4 in both fixes               |

### DIFF
--- a/benchmark/ops/bench_grouped_gemm_gmm.py
+++ b/benchmark/ops/bench_grouped_gemm_gmm.py
@@ -1,0 +1,205 @@
+###############################################################################
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""GMM (grouped_gemm) BF16 Grouped GEMM Baseline Benchmark.
+
+Reference: https://github.com/tgale96/grouped_gemm
+"""
+
+import argparse
+from datetime import datetime
+
+import grouped_gemm.ops as gmm_ops
+import pandas as pd
+import torch
+import torch.utils.benchmark as benchmark
+from config import (
+    check_allclose,
+    gen_grouped_gemm_group_lens,
+    gen_grouped_gemm_test_cases,
+    get_platform_info,
+    grouped_gemm_ref,
+)
+from tabulate import tabulate
+
+
+def check_grouped_gemm_gmm_correctness(a, b, batch_sizes, grad_out, dtype):
+    """Check correctness of gmm forward and backward against reference."""
+    out = gmm_ops.gmm(a, b, batch_sizes, trans_b=True)
+
+    group_lens = batch_sizes.to(device=a.device, dtype=torch.int64)
+    out_ref = grouped_gemm_ref(a.detach(), b.detach(), group_lens, trans_b=True)
+    fwd_correct = check_allclose(out.detach(), out_ref, dtype)
+
+    a_ref = a.detach().clone().requires_grad_()
+    b_ref = b.detach().clone().requires_grad_()
+    out_ref = grouped_gemm_ref(a_ref, b_ref, group_lens, trans_b=True)
+    out_ref.backward(grad_out)
+    out.backward(grad_out, retain_graph=True)
+    bwd_correct = check_allclose(a.grad, a_ref.grad, dtype)
+
+    a.grad = None
+
+    correct = fwd_correct and bwd_correct
+    status = "PASS" if correct else "FAIL"
+    print(f"Correctness Check: {status} (fwd={fwd_correct}, bwd={bwd_correct})")
+
+    return correct
+
+
+def profile_grouped_gemm_gmm(B, M, N, K, dtype, balance=True, num_topk=None):
+    """Profile BF16 Grouped GEMM using grouped_gemm (gmm) library."""
+    device = "cuda"
+    # batch_sizes must be on CPU for gmm
+    batch_sizes = gen_grouped_gemm_group_lens(B, M, balance=balance, num_topk=num_topk)
+    total_m = batch_sizes.sum().item()
+
+    a = torch.randn((total_m, K), dtype=dtype, device=device, requires_grad=True)
+    b = torch.randn((B, N, K), dtype=dtype, device=device, requires_grad=True)
+
+    out = gmm_ops.gmm(a, b, batch_sizes, trans_b=True)
+    grad_out = torch.randn_like(out)
+
+    correct = check_grouped_gemm_gmm_correctness(a, b, batch_sizes, grad_out, dtype)
+
+    def fwd_func():
+        return gmm_ops.gmm(a, b, batch_sizes, trans_b=True)
+
+    def fwd_bwd_func():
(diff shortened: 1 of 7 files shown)


xiaobochen-amd:
https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm.py#L53

Previously, some bad cases were skipped in unit tests—you can re-enable them and try again.

csrc/pytorch/grouped_gemm/hipblaslt_grouped_gemm.cpp
```diff
@@ -151,8 +154,11 @@ at::Tensor hipblaslt_grouped_gemm_fp8(at::Tensor &a, at::Tensor &b, at::Tensor &
         c               = at::empty({m, n}, a.options().dtype(out_dtype));
     }
 
-    const int64_t workspace_size = primus_turbo::get_hipblaslt_grouped_gemm_workspace_size();
-    at::Tensor    workspace =
+    const int64_t per_expert_workspace_size =
+        group_lens.numel() * primus_turbo::get_hipblaslt_workspace_size_in_byte();
+    const int64_t workspace_size = std::max<int64_t>(
```
RuibinCheung: I don't think it's a hang root cause.  In multi-stream grouped gemm, the gemm kernel in same stream was executed sequentially so `workspace[idx % stream]` will not cause race-condition. Here are some implementation reference of other repo for you:
TE：https://github.com/ROCm/TransformerEngine/blob/789035adbd08bd0ca040b38617e8e61680f5a7d9/transformer_engine/jax/cpp_extensions/gemm.py#L1495
legacy gg: https://github.com/AMD-AGI/grouped-gemm-ck/blob/e3b51851bec6c595fb63ad107aeac521d2171e00/csrc/grouped_gemm_hipblaslt.hip#L226-L228

RuibinCheung (changes requested):

csrc/pytorch/grouped_gemm/hipblaslt_grouped_gemm.cpp
kyle-256: Thanks for the comment! I am digging into it now.

kyle-256: the root cause is complicated, you can refer to my comment.

RuibinCheung (dismissed):
LGTM

kyle-256:
> https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm.py#L53
> 
> Previously, some bad cases were skipped in unit tests—you can re-enable them and try again.


bad cases can pass the unit tests

xiaobochen-amd:
> > https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm.py#L53
> > Previously, some bad cases were skipped in unit tests—you can re-enable them and try again.
> 
> bad cases can pass the unit tests

@kyle-256  https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm_fp8.py#L408

kyle-256:
> > > https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm.py#L53
> > > Previously, some bad cases were skipped in unit tests—you can re-enable them and try again.
> > 
> > 
> > bad cases can pass the unit tests
> 
> @kyle-256 https://github.com/AMD-AGI/Primus-Turbo/blob/main/tests/pytorch/ops/test_grouped_gemm_fp8.py#L408

fp8 related multistream bug on Mi300 is fixed in the latest commit. Mi355 is also tested and fixed.

PR #330, opened 2026-05-07
[WIP] [Feature] Add Turbo MXFP8 Grouped GEMM (gfx950) for MoE

kyle-256:
# Description

Add a turbo MXFP8 grouped GEMM backend (gfx950 / MI350) for MoE FFN
workloads.  Forward, dgrad and wgrad all run on hand-tuned 256x256x128
persistent kernels with E8M0 scales preshuffled into 16x4 col-major
blocks.  The wrapper handles `trans_b=False` and unaligned per-group
M_g transparently (zero-pad along M), so the new path passes the full
tensorwise parametrization (HYBRID excluded for now).

End-to-end overhead vs `tensorwise + Triton` on gpt-oss-20B shapes:
median **+12.8%**, with M=4096 cases matching or beating the tensorwise
baseline.

## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Turbo MXFP8 grouped GEMM kernels** (`csrc/kernels/grouped_gemm/`):
  fwd / dgrad share a 256x256x128 persistent kernel with optional
  store-side unpad via `c_group_offs`; wgrad uses a variable-K (M_g
  reduction) kernel.  uint32 BufferSRD num_records is clamped to avoid
  the >4 GB silent-truncation that masks valid LDGs as OOB.
- **Chunked-LDS dual-tensor preshuffle** for E8M0 scales: bit-identical
  to the single-GEMM kernel but stages each 16-row tile through LDS in
  256-byte column slabs, restoring fully coalesced HBM traffic on the
  fwd / dgrad / wgrad hot paths.
- **Per-group dual MXFP8 quant** (`quantize_mxfp8_dual_perg`): writes
  rowwise `(G, M, N_pad)` and colwise `(G, N, M_pad)` directly,
  eliminating the torch transpose+contiguous regroup that previously
  bottlenecked the dgrad B-side.
- **Fused per-group M zero-pad quant** (`quantize_mxfp8_dual_grouped`):
  computes the padded layout on GPU (no D2H sync) and materialises the
  padded layout straight into the output tensors, allocating outputs
  at the host-known upper bound so the call is fully async.
- **Race-freeness**: memory clobbers on MFMA / `clobber_*gpr` /
  `zero_agpr` / `read_agpr` inline asm + `clobber_vgpr_one` after
  `ds_read_b*`, matching the single-GEMM kernel's race profile.
- **Python wrapper** (`GroupedGemmFP8MXFunc`): handles `trans_b=False`
  (NT materialisation + grad transpose-back) and unaligned per-group
  M_g (128-aligned per-group zero-pad on A / grad_out) transparently.
- **Tests**: `test_grouped_gemm_fp8_mx_blockwise` mirrors the full
  tensorwise param sweep (`B`, `M`, `NK`, `ori_dtype`, `format`,
  `trans_b`, `balance`); HYBRID format is intentionally excluded.
- **Bench**: `bench_grouped_gemm_turbo.py` adds an `mxfp8` granularity
  for direct comparison with `tensorwise + Triton`.

## Checklist

# MX Grouped GEMM Bench

GPU: MI355 (gfx950) · dtype: bfloat16 · MX_BLOCKWISE (E4M3 + E8M0, block=32) ·
NT layout · CUDA-event timed (warmup 20 / iters 100). All numbers are
**TFLOPS** (higher is better).

- **Kernel-only** (`Fwd / Dgrad / Wgrad`): pure `turbo_grouped_gemm_fp8` /
  `turbo_grouped_gemm_variable_k_fp8`, inputs pre-quantized outside the
  timed region. Includes the in-kernel scale preshuffle.
  FLOPs = `2·B·M·N·K` per kernel.
- **End-to-end** (`Fwd+Q / Bwd+Q`): full Python op times.
  `Fwd+Q` = `grouped_gemm_fp8(...)` (A grouped quant + B per-group quant +
  GEMM, FLOPs = `2·B·M·N·K`).
  `Bwd+Q` = `out.sum().backward()` only (grad_out grouped quant + dgrad +
  wgrad, FLOPs = `4·B·M·N·K`).
- `bal` = balanced (`M_g = M`); `unb` = random per-group split, total = `B·M`.

### Kernel-only · Balanced

| Layer  |  B |    M |    N |    K |    Fwd |  Dgrad |  Wgrad |
|--------|---:|-----:|-----:|-----:|-------:|-------:|-------:|
| GateUp | 16 | 2048 | 4096 | 7168 | 2195.9 | 2093.6 | 1720.2 |
| GateUp | 16 | 4096 | 4096 | 7168 | 2450.1 | 2260.6 | 2186.3 |
| GateUp | 32 | 2048 | 4096 | 7168 | 2224.4 | 2116.5 | 1673.0 |
| GateUp | 32 | 4096 | 4096 | 7168 | 2456.4 | 2268.5 | 2048.4 |
| Down   | 16 | 2048 | 7168 | 2048 | 1645.0 | 2149.1 | 1708.8 |
| Down   | 16 | 4096 | 7168 | 2048 | 1784.5 | 2303.4 | 2110.4 |
| Down   | 32 | 2048 | 7168 | 2048 | 1679.1 | 2186.4 | 1695.4 |
| Down   | 32 | 4096 | 7168 | 2048 | 1741.7 | 2384.1 | 1996.6 |

### Kernel-only · Unbalanced

| Layer  |  B |    M |    N |    K |    Fwd |  Dgrad |  Wgrad |
|--------|---:|-----:|-----:|-----:|-------:|-------:|-------:|
| GateUp | 16 | 2048 | 4096 | 7168 | 1702.6 | 1761.1 | 1724.1 |
| GateUp | 16 | 4096 | 4096 | 7168 | 1780.8 | 1871.0 | 2104.5 |
| GateUp | 32 | 2048 | 4096 | 7168 | 1658.2 | 1712.5 | 1659.7 |
| GateUp | 32 | 4096 | 4096 | 7168 | 1610.1 | 1569.3 | 2002.7 |
| Down   | 16 | 2048 | 7168 | 2048 | 1488.5 | 1529.3 | 1701.3 |
| Down   | 16 | 4096 | 7168 | 2048 | 1606.1 | 1679.6 | 2045.7 |
| Down   | 32 | 2048 | 7168 | 2048 | 1554.4 | 1532.3 | 1677.5 |
| Down   | 32 | 4096 | 7168 | 2048 | 1447.4 | 1521.9 | 1964.5 |

### With Quant · Balanced

| Layer  |  B |    M |    N |    K |  Fwd+Q |  Bwd+Q |
|--------|---:|-----:|-----:|-----:|-------:|-------:|
| GateUp | 16 | 2048 | 4096 | 7168 | 1087.8 | 1263.6 |
| GateUp | 16 | 4096 | 4096 | 7168 | 1324.8 | 1573.4 |
| GateUp | 32 | 2048 | 4096 | 7168 | 1074.1 | 1228.7 |
| GateUp | 32 | 4096 | 4096 | 7168 | 1356.4 | 1523.2 |
| Down   | 16 | 2048 | 7168 | 2048 |  981.3 | 1151.2 |
| Down   | 16 | 4096 | 7168 | 2048 | 1186.3 | 1400.2 |
| Down   | 32 | 2048 | 7168 | 2048 | 1009.8 | 1109.5 |
| Down   | 32 | 4096 | 7168 | 2048 | 1211.3 | 1312.9 |

### With Quant · Unbalanced

| Layer  |  B |    M |    N |    K |  Fwd+Q |  Bwd+Q |
|--------|---:|-----:|-----:|-----:|-------:|-------:|
| GateUp | 16 | 2048 | 4096 | 7168 |  955.4 | 1213.9 |
| GateUp | 16 | 4096 | 4096 | 7168 | 1140.4 | 1435.4 |
| GateUp | 32 | 2048 | 4096 | 7168 |  913.3 | 1160.2 |
| GateUp | 32 | 4096 | 4096 | 7168 | 1014.0 | 1320.2 |
| Down   | 16 | 2048 | 7168 | 2048 |  953.9 | 1040.5 |
| Down   | 16 | 4096 | 7168 | 2048 | 1163.4 | 1243.3 |
| Down   | 32 | 2048 | 7168 | 2048 |  957.2 |  996.8 |
| Down   | 32 | 4096 | 7168 | 2048 | 1073.6 | 1138.3 |

### Kernel-only · Balanced

| Layer  |  B |    M |    N |    K |    Fwd |  Dgrad |  Wgrad |
|--------|---:|-----:|-----:|-----:|-------:|-------:|-------:|
| GateUp |  4 | 2048 | 5760 | 2880 | 1682.6 | 1691.3 | 1353.1 |
| GateUp |  4 | 4096 | 5760 | 2880 | 1938.1 | 2214.2 | 1745.3 |
| GateUp | 32 | 2048 | 5760 | 2880 | 1867.1 | 2094.3 | 1531.8 |
| GateUp | 32 | 4096 | 5760 | 2880 | 1982.7 | 2262.1 | 1814.0 |
| Down   |  4 | 2048 | 2880 | 2880 | 1252.7 | 1293.0 | 1159.2 |
| Down   |  4 | 4096 | 2880 | 2880 | 1781.9 | 1778.5 | 1521.1 |
| Down   | 32 | 2048 | 2880 | 2880 | 1745.6 | 1713.8 | 1539.9 |
| Down   | 32 | 4096 | 2880 | 2880 | 1900.6 | 1892.3 | 1829.1 |

### Kernel-only · Unbalanced

| Layer  |  B |    M |    N |    K |    Fwd |  Dgrad |  Wgrad |
|--------|---:|-----:|-----:|-----:|-------:|-------:|-------:|
| GateUp |  4 | 2048 | 5760 | 2880 | 1165.7 | 1215.5 | 1341.4 |
| GateUp |  4 | 4096 | 5760 | 2880 | 1325.3 | 1552.5 | 1686.5 |
| GateUp | 32 | 2048 | 5760 | 2880 | 1430.2 | 1453.1 | 1493.0 |
| GateUp | 32 | 4096 | 5760 | 2880 | 1458.5 | 1463.6 | 1755.3 |
| Down   |  4 | 2048 | 2880 | 2880 |  947.0 |  960.2 | 1111.1 |
| Down   |  4 | 4096 | 2880 | 2880 | 1267.1 | 1276.8 | 1436.3 |
| Down   | 32 | 2048 | 2880 | 2880 | 1322.5 | 1336.9 | 1494.6 |
| Down   | 32 | 4096 | 2880 | 2880 | 1337.9 | 1332.0 | 1703.4 |

### With Quant · Balanced

| Layer  |  B |    M |    N |    K |  Fwd+Q |  Bwd+Q |
|--------|---:|-----:|-----:|-----:|-------:|-------:|
| GateUp |  4 | 2048 | 5760 | 2880 |  820.2 |  986.2 |
| GateUp |  4 | 4096 | 5760 | 2880 | 1101.1 | 1292.5 |
| GateUp | 32 | 2048 | 5760 | 2880 |  977.7 | 1158.8 |
| GateUp | 32 | 4096 | 5760 | 2880 | 1211.7 | 1337.6 |
| Down   |  4 | 2048 | 2880 | 2880 |  592.5 |  804.5 |
| Down   |  4 | 4096 | 2880 | 2880 |  882.7 | 1083.4 |
| Down   | 32 | 2048 | 2880 | 2880 |  868.3 | 1047.4 |
| Down   | 32 | 4096 | 2880 | 2880 | 1068.7 | 1236.1 |

### With Quant · Unbalanced

| Layer  |  B |    M |    N |    K |  Fwd+Q |  Bwd+Q |
|--------|---:|-----:|-----:|-----:|-------:|-------:|
| GateUp |  4 | 2048 | 5760 | 2880 |  684.2 |  907.0 |
| GateUp |  4 | 4096 | 5760 | 2880 |  876.8 | 1155.0 |
| GateUp | 32 | 2048 | 5760 | 2880 |  840.1 | 1014.5 |
| GateUp | 32 | 4096 | 5760 | 2880 |  995.2 | 1142.2 |
| Down   |  4 | 2048 | 2880 | 2880 |  509.3 |  734.3 |
| Down   |  4 | 4096 | 2880 | 2880 |  737.5 |  961.3 |
| Down   | 32 | 2048 | 2880 | 2880 |  754.0 |  955.1 |
| Down   | 32 | 4096 | 2880 | 2880 |  870.4 | 1068.4 |

### DIFF
--- a/csrc/include/primus_turbo/device/register.cuh
+++ b/csrc/include/primus_turbo/device/register.cuh
@@ -62,6 +62,7 @@ template <int AGPR> __device__ __forceinline__ void clobber_agpr_one() {
 
 template <int AC> __device__ __forceinline__ void zero_agpr() {
     asm volatile("v_accvgpr_write_b32 a[%0], 0" : : "n"(AC));
+    clobber_agpr_one<AC>();
 }
 
 // Read N consecutive AGPRs starting at AC, returned as type T.
--- a/csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
+++ b/csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
@@ -1,0 +1,488 @@
+// Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+//
+// See LICENSE for license information.
+
+#pragma once
+
+// Reuse GEMM_Tile_MXFP8_NT_*, preshuffle_scale_16x4_kernel, and BufferSRD helpers
+// from the single-GEMM kernel.
+#include "../../gemm/turbo/turbo_gemm_mxfp8_kernel.h"
+#include "primus_turbo/quantization.h"
+
+namespace primus_turbo {
+namespace turbo {
+
+// Per-tile compute body for the persistent kernel.  Computes one 256x256
+// output tile for (pid_m_local, pid_n_idx) within a group whose per-group
+// base pointers (a_grp_ptr/b_grp_ptr/.../c_grp_ptr) and per-group bounds
+// (M_g real for output, M_g_in padded for input/scale SRD) the caller
+// resolved in the outer kernel.  Inner takes only resolved pointers so
+// the per-group i64 offset arithmetic stays out of the inner kernel's
+// VGPR live set — keeps general-purpose VGPR (v[0:111]) under the cap
+// imposed by the pinned data/scale buffers (v[112:255]).
+template <typename GemmTile, typename AType, typename BType, typename CType>
+__device__ __forceinline__ void turbo_grouped_gemm_mxfp8_compute_tile(
+    GemmTile &tile, typename GemmTile::ASmemSubtile (*a_smem_tile)[4],
+    typename GemmTile::BSmemSubtile (*b_smem_tile)[4],
+    typename GemmTile::AScaleSmemSubtile (*a_s_smem_tile)[4],
+    typename GemmTile::BScaleSmemSubtile (*b_s_smem_tile)[4], const AType *a_grp_ptr,
+    const BType *b_grp_ptr, const uint32_t *a_s_grp_ptr, const uint32_t *b_s_grp_ptr,
+    CType *c_grp_ptr, const int32_t pid_m_local, const int32_t pid_n_local, const int32_t M_g,
+    const int32_t M_g_in, const uint32_t n, const uint32_t k) {
+    const int32_t pid_m = pid_m_local * 256;
+    const int32_t pid_n = pid_n_local * 256;
+    if (pid_m >= M_g || pid_n >= (int32_t) n)
+        return;
+
+    tile.reserve_pinned_regs();
+
+    const uint32_t lane_id = tile.lane_id;
+    const uint32_t warp_id = tile.warp_id;
+    const uint32_t warp_m  = tile.warp_m;
+    const uint32_t warp_n  = tile.warp_n;
+
+    const AType    *a_base_ptr   = a_grp_ptr + (int64_t) pid_m * k;
+    const BType    *b_base_ptr   = b_grp_ptr + (int64_t) pid_n * k;
+    const uint32_t  scale_cols   = (k + GemmTile::MX_BLOCK_SIZE - 1) / GemmTile::MX_BLOCK_SIZE;
+    const uint32_t *a_s_base_ptr = a_s_grp_ptr + (int64_t) pid_m * scale_cols;
+    const uint32_t *b_s_base_ptr = b_s_grp_ptr + (int64_t) pid_n * scale_cols;
+
+    uint32_t ldg_offsets[2];
+    tile.compute_ldg_offsets(ldg_offsets, k);
+    uint32_t lds_offsets[2];
+    tile.compute_lds_offsets(lds_offsets);
+    const uint32_t scale_ldg_offset = lane_id;
+    const uint32_t scale_sts_offset = lane_id;
+    const uint32_t scale_lds_offset = lane_id;
+
+    const uint32_t  a_remaining  = ((uint32_t) M_g_in - pid_m) * k * sizeof(AType);
+    const uint32_t  b_remaining  = (n - pid_n) * k * sizeof(BType);
+    const uint32_t  as_remaining = ((uint32_t) M_g_in - pid_m) * scale_cols * sizeof(uint32_t);
+    const uint32_t  bs_remaining = (n - pid_n) * scale_cols * sizeof(uint32_t);
+    const BufferSRD a_srd(a_base_ptr, a_remaining);
+    const BufferSRD b_srd(b_base_ptr, b_remaining);
+    const BufferSRD a_s_srd(a_s_base_ptr, as_remaining);
+    const BufferSRD b_s_srd(b_s_base_ptr, bs_remaining);
+
+    constexpr int32_t DATA_STRIDE  = GemmTile::BLOCK_SIZE_K;
+    constexpr int32_t SCALE_STRIDE = GemmTile::SCALE_FRAG_SIZE * sizeof(uint32_t);
+
+    // ── Load tile 0 → smem[0], tile 1 → smem[1] ──
+    tile.template load_a_gmem_to_smem_half_srd<0>(a_srd, ldg_offsets, a_smem_tile[0]);
+    tile.template load_a_gmem_to_smem_half_srd<1>(a_srd, ldg_offsets, a_smem_tile[0]);
+    tile.template load_b_gmem_to_smem_half_srd<0>(b_srd, ldg_offsets, b_smem_tile[0]);
+    tile.template load_b_gmem_to_smem_half_srd<1>(b_srd, ldg_offsets, b_smem_tile[0]);
+    tile.template load_a_scale_gmem_to_smem_half_srd<0>(a_s_srd, scale_ldg_offset, a_s_smem_tile[0],
+                                                        scale_sts_offset, scale_cols);
+    tile.template load_a_scale_gmem_to_smem_half_srd<1>(a_s_srd, scale_ldg_offset, a_s_smem_tile[0],
+                                                        scale_sts_offset, scale_cols);
+    tile.template load_b_scale_gmem_to_smem_half_srd<0>(b_s_srd, scale_ldg_offset, b_s_smem_tile[0],
+                                                        scale_sts_offset, scale_cols);
+    tile.template load_b_scale_gmem_to_smem_half_srd<1>(b_s_srd, scale_ldg_offset, b_s_smem_tile[0],
+                                                        scale_sts_offset, scale_cols);
+
+    tile.template load_a_gmem_to_smem_half_srd<0>(a_srd, ldg_offsets, a_smem_tile[1], DATA_STRIDE);
+    tile.template load_a_gmem_to_smem_half_srd<1>(a_srd, ldg_offsets, a_smem_tile[1], DATA_STRIDE);
+    tile.template load_b_gmem_to_smem_half_srd<0>(b_srd, ldg_offsets, b_smem_tile[1], DATA_STRIDE);
+    tile.template load_b_gmem_to_smem_half_srd<1>(b_srd, ldg_offsets, b_smem_tile[1], DATA_STRIDE);
+    tile.template load_a_scale_gmem_to_smem_half_srd<0>(a_s_srd, scale_ldg_offset, a_s_smem_tile[1],
+                                                        scale_sts_offset, scale_cols, SCALE_STRIDE);
+    tile.template load_a_scale_gmem_to_smem_half_srd<1>(a_s_srd, scale_ldg_offset, a_s_smem_tile[1],
+                                                        scale_sts_offset, scale_cols, SCALE_STRIDE);
+    tile.template load_b_scale_gmem_to_smem_half_srd<0>(b_s_srd, scale_ldg_offset, b_s_smem_tile[1],
+                                                        scale_sts_offset, scale_cols, SCALE_STRIDE);
+    tile.template load_b_scale_gmem_to_smem_half_srd<1>(b_s_srd, scale_ldg_offset, b_s_smem_tile[1],
+                                                        scale_sts_offset, scale_cols, SCALE_STRIDE);
+
+    tile.zero_c_agpr();
+    wait_vmcnt<0>();
+    // Drain in-flight buffer_load_lds GMEM->LDS writes before the barrier.
+    // (Replaces a per-tile __threadfence(); L2 coherence is hoisted to kernel entry.)
+    wait_lgkmcnt<0>();
+    __builtin_amdgcn_s_barrier();
+
+    uint32_t       cur     = 0;
+    uint32_t       next    = 1;
+    const uint32_t k_iters = (k + GemmTile::BLOCK_SIZE_K - 1) / GemmTile::BLOCK_SIZE_K;
+
+    // ── Prologue: issue LDS for A0/B0 ──
+    GemmTile::template load_data_subtile_pinned<GemmTile::PIN_A0>(
+        a_smem_tile[cur][warp_m].u32_ptr(), lds_offsets);
+    GemmTile::template load_scale_subtile_pinned<GemmTile::PIN_AS0>(
+        a_s_smem_tile[cur][warp_m].u32_ptr(), scale_lds_offset);
+    GemmTile::template load_data_subtile_pinned<GemmTile::PIN_B0>(
+        b_smem_tile[cur][warp_n].u32_ptr(), lds_offsets);
+    GemmTile::template load_scale_subtile_pinned<GemmTile::PIN_BS0>(
+        b_s_smem_tile[cur][warp_n].u32_ptr(), scale_lds_offset);
+
+    // Drain prologue ds_reads before LDG below — both target cur[0,1].
+    wait_lgkmcnt<0>();
+    if (k_iters > 2) {
+        tile.template load_a_gmem_to_smem_half_srd<0>(a_srd, ldg_offsets, a_smem_tile[cur],
+                                                      2 * DATA_STRIDE);
+        tile.template load_a_scale_gmem_to_smem_half_srd<0>(a_s_srd, scale_ldg_offset,
+                                                            a_s_smem_tile[cur], scale_sts_offset,
+                                                            scale_cols, 2 * SCALE_STRIDE);
+        tile.template load_b_gmem_to_smem_half_srd<0>(b_srd, ldg_offsets, b_smem_tile[cur],
+                                                      2 * DATA_STRIDE);
+        tile.template load_b_scale_gmem_to_smem_half_srd<0>(b_s_srd, scale_ldg_offset,
+                                                            b_s_smem_tile[cur], scale_sts_offset,
+                                                            scale_cols, 2 * SCALE_STRIDE);
+    }
+    wait_lgkmcnt<0>();
+
+    int32_t base_data_soff[4], base_scale_soff[4];
+    tile.precompute_base_soff(base_data_soff, base_scale_soff, scale_cols);
+
+    const uint32_t sts_wb =
+        __builtin_amdgcn_readfirstlane(warp_id * GemmTile::MFMA_SIZE_M * GemmTile::MFMA_SIZE_K);
+    const uint32_t s_smem_off = __builtin_amdgcn_readfirstlane((warp_id * 64 + scale_sts_offset) *
+                                                               (uint32_t) sizeof(uint32_t));
+    const uint32_t scale_gmem_byte_off = scale_ldg_offset * (uint32_t) sizeof(uint32_t);
+
+    // Re-assert pinned reservation right before main MFMA loop entry.
+    tile.reserve_pinned_regs();
+
+    // ── Main loop ──
+    const uint32_t main_iters = k_iters > 3 ? k_iters - 3 : 0;
+    int32_t        data_off = 2 * DATA_STRIDE, scale_off = 2 * SCALE_STRIDE;
+    for (uint32_t ki = 0; ki < main_iters;
+         ++ki, data_off += DATA_STRIDE, scale_off += SCALE_STRIDE) {
+        const int32_t next_data_off  = data_off + DATA_STRIDE;
+        const int32_t next_scale_off = scale_off + SCALE_STRIDE;
+
+        // Phase 1: MFMA A0×B0, LDG B1→cur[2,3]
+        {
+            uint32_t dm0_0 = __builtin_amdgcn_readfirstlane(b_smem_tile[cur][2].u32_ptr()) + sts_wb;
+            uint32_t dm0_1 = __builtin_amdgcn_readfirstlane(b_smem_tile[cur][3].u32_ptr()) + sts_wb;
+            uint32_t sm0_0 =
+                __builtin_amdgcn_readfirstlane(b_s_smem_tile[cur][2].u32_ptr()) + s_smem_off;
+            uint32_t sm0_1 =
+                __builtin_amdgcn_readfirstlane(b_s_smem_tile[cur][3].u32_ptr()) + s_smem_off;
+            GemmTile::template phase_mfma_lds_ldg<GemmTile::PIN_A0, GemmTile::PIN_AS0,
+                                                  GemmTile::PIN_B0, GemmTile::PIN_BS0, 0, 0,
+                                                  GemmTile::PIN_B1, GemmTile::PIN_BS1>(
+                b_smem_tile[cur][warp_n + 2].u32_ptr(), lds_offsets,
+                b_s_smem_tile[cur][warp_n + 2].u32_ptr(), scale_lds_offset, b_srd, ldg_offsets,
+                dm0_0, dm0_1, base_data_soff[2] + data_off, base_data_soff[3] + data_off, b_s_srd,
+                scale_gmem_byte_off, sm0_0, sm0_1, base_scale_soff[2] + scale_off,
+                base_scale_soff[3] + scale_off);
+        }
+
+        // Phase 2: MFMA A0×B1, LDG A1→cur[2,3]
+        {
+            uint32_t dm0_0 = __builtin_amdgcn_readfirstlane(a_smem_tile[cur][2].u32_ptr()) + sts_wb;
+            uint32_t dm0_1 = __builtin_amdgcn_readfirstlane(a_smem_tile[cur][3].u32_ptr()) + sts_wb;
+            uint32_t sm0_0 =
+                __builtin_amdgcn_readfirstlane(a_s_smem_tile[cur][2].u32_ptr()) + s_smem_off;
+            uint32_t sm0_1 =
+                __builtin_amdgcn_readfirstlane(a_s_smem_tile[cur][3].u32_ptr()) + s_smem_off;
+            GemmTile::template phase_mfma_lds_ldg<GemmTile::PIN_A0, GemmTile::PIN_AS0,
+                                                  GemmTile::PIN_B1, GemmTile::PIN_BS1, 0, 1,
+                                                  GemmTile::PIN_A1, GemmTile::PIN_AS1>(
+                a_smem_tile[cur][warp_m + 2].u32_ptr(), lds_offsets,
+                a_s_smem_tile[cur][warp_m + 2].u32_ptr(), scale_lds_offset, a_srd, ldg_offsets,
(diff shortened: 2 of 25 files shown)


csrc/include/primus_turbo/device/memory.cuh
```diff
@@ -113,6 +115,14 @@ __device__ __forceinline__ void ds_read_pinned(uint32_t lds_addr) {
...
         asm volatile("ds_read_b128 v[%0:%1], %2 offset:%3"
             : : "n"(VDST), "n"(VDST + 3), "v"(lds_addr), "n"(IMM_OFFSET) : "memory");
+
+    clobber_vgpr_one<VDST>();
+    if constexpr (Bytes >= 8)
+        clobber_vgpr_one<VDST + 1>();
+    if constexpr (Bytes == 16) {
+        clobber_vgpr_one<VDST + 2>();
+        clobber_vgpr_one<VDST + 3>();
+    }
```
xiaobochen-amd: Why do we need to add clobber_vgpr_one here for ds_read? The related registers should already be clobbered before the ds_read.

csrc/include/primus_turbo/device/mfma.cuh
```diff
@@ -58,16 +58,16 @@ template <typename AType, typename BType> struct mfma_scale_f32_16x16x128_f8f6f4
...
+                : : "n"(PIN_ACC), "n"(PIN_ACC + 3), "n"(PIN_A), "n"(PIN_A + 7), "n"(PIN_B), "n"(PIN_B + 7), "n"(PIN_SA), "n"(PIN_SB) : "memory");
         else if constexpr (cbsz == 0 && blgp == 1)
             asm volatile("v_mfma_scale_f32_16x16x128_f8f6f4 v[%0:%1], v[%2:%3], v[%4:%5], v[%0:%1], v[%6], v[%7] op_sel_hi:[0,0,0] blgp:1"
-                : : "n"(PIN_ACC), "n"(PIN_ACC + 3), "n"(PIN_A), "n"(PIN_A + 7), "n"(PIN_B), "n"(PIN_B + 7), "n"(PIN_SA), "n"(PIN_SB));
+                : : "n"(PIN_ACC), "n"(PIN_ACC + 3), "n"(PIN_A), "n"(PIN_A + 7), "n"(PIN_B), "n"(PIN_B + 7), "n"(PIN_SA), "n"(PIN_SB) : "memory");
         else
             asm volatile("v_mfma_scale_f32_16x16x128_f8f6f4 v[%0:%1], v[%2:%3], v[%4:%5], v[%0:%1], v[%6], v[%7] op_sel_hi:[0,0,0] cbsz:1 blgp:1"
-                : : "n"(PIN_ACC), "n"(PIN_ACC + 3), "n"(PIN_A), "n"(PIN_A + 7), "n"(PIN_B), "n"(PIN_B + 7), "n"(PIN_SA), "n"(PIN_SB));
+                : : "n"(PIN_ACC), "n"(PIN_ACC + 3), "n"(PIN_A), "n"(PIN_A + 7), "n"(PIN_B), "n"(PIN_B + 7), "n"(PIN_SA), "n"(PIN_SB) : "memory");
```
xiaobochen-amd: v_mfma_scale_f32_16x16x128_f8f6f4 only modifies registers, so I don't think we need the `"memory"` here.

csrc/include/primus_turbo/device/register.cuh
```diff
@@ -21,7 +21,7 @@ namespace primus_turbo::device {
 template <int AGPR> __device__ __forceinline__ void clobber_agpr_one() {
     static_assert(AGPR >= 0 && AGPR <= 255, "AGPR must be in [0, 255]");
     // clang-format off
-#define CLOBBER_AREG_CASE(N) case N: asm volatile("" ::: "a" #N); break;
+#define CLOBBER_AREG_CASE(N) case N: asm volatile("" ::: "a" #N, "memory"); break;
```
xiaobochen-amd: Same as above.

csrc/include/primus_turbo/device/register.cuh
```diff
@@ -61,7 +61,8 @@ template <int AGPR> __device__ __forceinline__ void clobber_agpr_one() {
 }
 
 template <int AC> __device__ __forceinline__ void zero_agpr() {
-    asm volatile("v_accvgpr_write_b32 a[%0], 0" : : "n"(AC));
+    asm volatile("v_accvgpr_write_b32 a[%0], 0" : : "n"(AC) : "memory");
+    clobber_agpr_one<AC>();
```
xiaobochen-amd, line 65: Same as above.

csrc/include/primus_turbo/device/register.cuh
```diff
@@ -72,22 +73,22 @@ template <typename T, int AC> __device__ __forceinline__ T read_agpr() {
...
+    if constexpr (N >=  7) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[6])  : "n"(AC + 6) : "memory");
+    if constexpr (N >=  8) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[7])  : "n"(AC + 7) : "memory");
+    if constexpr (N >=  9) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[8])  : "n"(AC + 8) : "memory");
+    if constexpr (N >= 10) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[9])  : "n"(AC + 9) : "memory");
+    if constexpr (N >= 11) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[10]) : "n"(AC + 10) : "memory");
+    if constexpr (N >= 12) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[11]) : "n"(AC + 11) : "memory");
+    if constexpr (N >= 13) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[12]) : "n"(AC + 12) : "memory");
+    if constexpr (N >= 14) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[13]) : "n"(AC + 13) : "memory");
+    if constexpr (N >= 15) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[14]) : "n"(AC + 14) : "memory");
+    if constexpr (N >= 16) asm volatile("v_accvgpr_read_b32 %0, a[%1]" : "=v"(raw[15]) : "n"(AC + 15) : "memory");
```
xiaobochen-amd: Same as above.

primus_turbo/pytorch/ops/grouped_gemm_fp8.py
```diff
@@ -49,6 +52,82 @@ def _ensure_contiguous_grad_out(grad_out: torch.Tensor) -> torch.Tensor:
     return grad_out if grad_out.is_contiguous() else grad_out.contiguous()
 
 
+def _mxfp8_grid_x_hint(total_M: int, G: int) -> int:
```
xiaobochen-amd: Both inputs are available C++-side as a.size(0) and group_lens.size(0). Moving this into the launcher lets the dispatcher path skip the group_lens.cpu() sync too.

csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
```diff
@@ -0,0 +1,491 @@
...
+// (the original/unpadded layout) instead of ``group_offs_ptr[group_id]``,
+// fusing the per-group-padding extract directly into the GEMM store.
+template <typename GemmTile, typename AType, typename BType, typename CType>
+__device__ __forceinline__ void turbo_grouped_gemm_mxfp8_compute_tile(
+    GemmTile &tile, typename GemmTile::ASmemSubtile (*a_smem_tile)[4],
+    typename GemmTile::BSmemSubtile (*b_smem_tile)[4],
+    typename GemmTile::AScaleSmemSubtile (*a_s_smem_tile)[4],
+    typename GemmTile::BScaleSmemSubtile (*b_s_smem_tile)[4], const AType *a_ptr,
+    const BType *b_ptr, const uint32_t *a_s_ptr, const uint32_t *b_s_ptr, CType *c_ptr,
+    const int64_t *group_offs_ptr, const int64_t *c_group_offs_ptr, const int32_t group_id,
```
xiaobochen-amd: Naming is asymmetric: group_offs_ptr is actually input-side (padded) offsets while c_group_offs_ptr is output-side (compact). Consider renaming to group_offs_in_ptr / group_offs_out_ptr (or a_group_offs_ptr / c_group_offs_ptr) so the asymmetry is visible at the callsite.

csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
```diff
@@ -0,0 +1,491 @@
...
+        const uint32_t c_offset_hi = __builtin_amdgcn_readfirstlane(
+            static_cast<uint32_t>(static_cast<uint64_t>(c_group_offs_ptr[group_id]) >> 32));
+        const int64_t c_group_offset =
+            (static_cast<int64_t>(c_offset_hi) << 32) | static_cast<int64_t>(c_offset_lo);
+        c_m_global = c_group_offset + (int64_t) pid_m;
+        // M_g_padded = ceil(M_g, 128).  Quant kernel pads each
+        // per-group region to 128-aligned, so this matches
+        // group_offs_padded[g+1] - group_offs_padded[g] without an
+        // extra scalar load.
+        M_g_in = (M_g + 127) & ~127;
```
xiaobochen-amd: 127 hardcode?

csrc/include/primus_turbo/device/register.cuh
```diff
@@ -119,7 +120,7 @@ template <int START, int END> __device__ __forceinline__ void zero_agpr_range()
 template <int VGPR> __device__ __forceinline__ void clobber_vgpr_one() {
     static_assert(VGPR >= 0 && VGPR <= 255, "VGPR must be in [0, 255]");
     // clang-format off
-#define CLOBBER_VREG_CASE(N) case N: asm volatile("" ::: "v" #N); break;
+#define CLOBBER_VREG_CASE(N) case N: asm volatile("" ::: "v" #N, "memory"); break;
```
xiaobochen-amd: Same as above.

csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
xiaobochen-amd: Just realized that in real workloads the inputs probably always need to be padded anyway. In that case, it may be cleaner to stop overloading the presence of c_group_offs_ptr as a mode switch, and instead always pass explicit a_group_offs_ptr, b_group_offs_ptr, and c_group_offs_ptr. The interface becomes much more straightforward.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp8_impl.py
```diff
@@ -515,8 +585,73 @@ def execute(
...
+        group_lens: torch.Tensor,
+        group_offs: torch.Tensor,
+        trans_a: bool,
+        trans_b: bool,
+        trans_c: bool,
+        out_dtype: torch.dtype,
+        granularity: ScalingGranularity,
+        num_cu: int | None,
+        **kwargs,
+    ) -> bool:
```
xiaobochen-amd, line 620: Need check shape?

primus_turbo/pytorch/ops/grouped_gemm_fp8.py
```diff
@@ -373,6 +452,163 @@ def backward(ctx, grad_out):
...
+            b_fp8_col,
+            b_scale_inv_col,
+            group_lens,
+            group_offs,
+            group_offs_padded,
+        )
+        ctx.config = config
+        ctx.out_dtype = out_dtype
+        ctx.num_cu = num_cu
+        ctx.trans_b_orig = trans_b
```
xiaobochen-amd: ctx.trans_b

csrc/include/primus_turbo/device/memory.cuh
kyle-256: Tried removing them — 100% race in grouped MXFP8 dgrad (max diff inf) and 0.7% race in dense MXFP8 mxgemm at 4096³, exactly matching the historical race profile we hit before.

The pre-ds_read `clobber_vgpr_one<>` reserves the dst VGPRs from being reused, but doesn't fence the compiler from sinking the consuming MFMA above the `ds_read`. The post-asm clobber on the dst VGPR is what creates that ordering edge.

csrc/include/primus_turbo/device/mfma.cuh
kyle-256: removed all "memory"

csrc/kernels/grouped_gemm/turbo/turbo_grouped_gemm_mxfp8_kernel.h
kyle-256: Done. Launcher now always passes a non-null `c_group_offs_ptr` (caller's compact-output offsets when supplied; `a_group_offs_ptr` as fallback for the in-place layout). Kernel takes a single straight-line extract path with no mode switch on the pointer.

On `b_group_offs_ptr`: B is laid out as `[G, N, K]` and indexed by `group_id` directly — no per-group row offsets for B, so no third pointer to add.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp8_impl.py
kyle-256: Done — added: `n_fwd % 16 == 0 and k_fwd % 16 == 0` (wgrad MFMA tile is 16×16, output axes must clear)

kyle-256:
@xiaobochen-amd Thanks for the commit! I have updated my code by the comments. Please review again, thanks.

csrc/include/primus_turbo/device/memory.cuh
xiaobochen-amd: 1. How was the 0.7% race obtained?

2. In the Turbo MXFP8 GEMM, there should be no race in any phase between MFMA and DS. MFMA computes the current block, while DS loads the data for the next block. Any cross-block race has already been constrained using __builtin_amdgcn_s_barrier.

xiaobochen-amd: The role of `clobber_vgpr_one` is to allocate/register VGPRs. It cannot prevent the compiler from using them.

xiaobochen-amd: Additionally, my personal understanding is that using clobber_vgpr_one on ds_read is not a semantically correct way to enforce ordering, because it only operates at the level of register allocation and does not involve modeling of memory dependencies or execution ordering. Any apparent “correctness effect” is likely just an indirect result of changes in compiler scheduling behavior, rather than a reliable ordering guarantee. Race conditions should instead be addressed using standard synchronization and memory-dependency mechanisms such as s_waitcnt and barriers.

kyle-256: It seems to be some bug that led to this result. No race in main branch. I will take time to find out.

kyle-256: The grouped GEMM race is triggered by compiler-emitted SGPR spill: the spill reload uses vmcnt, which gfx950 does not FIFO-order against buffer_load_lds, so a stale m0 can leak into the LDS write address. The latest commits both reduce spill (CK refactor + smem packing + -sink-insts-to-avoid-spills; private_segment drops 44 → 20 bytes) and move the prologue LDS commit onto lgkmcnt (2-step buffer_load → ds_write), decoupling it from the spill's vmcnt traffic — race goes to zero.

Tested on both gemm kernel and grouped gemm kernels, no race condition happened again.

xiaobochen-amd: Why do we need the two-step path? It copies the data through VGPR first and then into SMEM, right? If so, it seems to break the whole design philosophy here:

1. it breaks the counter separation model.

vmcnt is supposed to track only buffer_load_lds (GMEM→LDS)
lgkmcnt is supposed to track only ds_read_pinned (LDS→VGPR)

The two-step path introduces ds_write into lgkmcnt, which muddies the semantics of the two counters.

2. the two-step path is effectively a reverse optimization.

The original purpose of buffer_load_lds was zero-VGPR buffering plus fully asynchronous transfer. The two-step path regresses into GMEM→VGPR→LDS, which either consumes extra VGPRs or sacrifices asynchronicity.

3. SGPR resources on the hardware should already be extremely abundant.

xiaobochen-amd: “CK refactor” ？

kyle-256: The 2-step prologue path is removed, and VGPR / SGPR spill is fully eliminated — all kernel specs now show `sgpr_spill = 0 / vgpr_spill = 0 / private_segment = 0`, and `scratch_load` / `scratch_store` are completely gone from disasm. The vmcnt race root cause is fixed at the source.

But `clobber_vgpr_one` / `clobber_agpr_one` still need to stay:

- The spill race and the clobber race are two different things. Clobber prevents **compiler reordering** — `ds_read_pinned` and MFMA both use `"n"(VDST)` constant operands that encode the register name into the asm string, so the compiler sees no def-use chain. Per GCC: two `asm volatile` statements are not ordered with respect to each other. Without the clobbers, the compiler is free to reorder MFMA before the consuming `ds_read` → reads stale VGPR.
- Verified empirically: with spill at 0, removing both clobbers still races — grouped fwd 5.5%, wgrad 0.05%.

primus_turbo/pytorch/ops/grouped_gemm_fp8.py
```diff
@@ -49,6 +50,66 @@ def _ensure_contiguous_grad_out(grad_out: torch.Tensor) -> torch.Tensor:
     return grad_out if grad_out.is_contiguous() else grad_out.contiguous()
 
 
+def _turbo_grouped_gemm_mxfp8(
```
xiaobochen-amd: The code here does not align with our backend design

tests/pytorch/ops/test_grouped_gemm_fp8.py
```diff
@@ -385,6 +403,38 @@ def test_grouped_gemm_fp8_blockwise_triton_deterministic(
...
+@pytest.mark.parametrize("B", _DET_B_VALUES)
+@pytest.mark.parametrize("M", _DET_M_VALUES)
+@pytest.mark.parametrize("NK", _DET_NK_VALUES)
+@pytest.mark.parametrize("ori_dtype", ORI_DTYPE_VALUES)
+@pytest.mark.parametrize("format", FORMAT_VALUES)
+@pytest.mark.parametrize("trans_b", [True])
+@pytest.mark.parametrize("balance", [True, False])
+@pytest.mark.deterministic
+def test_grouped_gemm_fp8_mx_blockwise_deterministic(B, M, NK, ori_dtype, format, trans_b, balance):
+    if get_device_compute_capability() != (9, 5):
```
xiaobochen-amd: check_mxfp8_support

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp8_impl.py
```diff
@@ -647,6 +802,9 @@ def grouped_gemm_fp8_impl_meta(
     num_cu: int | None,
     default_backend: int,
     maybe_pre_sync: bool = False,
+    c_group_offs: torch.Tensor | None = None,
```
RuibinCheung: In mxfp8 gemm, we padded a and b in quantization kernel. So we assume a and b are padded in grouped gemm kernel. Please remove these unused arguments.

csrc/pytorch/quantization/quantization.cpp
```diff
@@ -569,6 +569,226 @@ quantize_mxfp8_dual(const at::Tensor input, const at::ScalarType dest_dtype,
...
+                           " for shuffled colwise FP8. But got N=", N);
+    }
+
+    auto device = input.device();
+    auto stream = at::cuda::getCurrentCUDAStream();
+
+    // Fold layout computation into this op (no separate D2H round-trip).
+    auto group_lens_padded = at::empty_like(group_lens);
+    auto group_offs_padded = at::empty({G + 1}, group_offs.options());
+    primus_turbo::compute_padded_layout_gpu(
```
RuibinCheung, line 757: The MXFP8_PADDING_ALIGN_SIZE means n dim padding align size. For group offs, we may want to pad m dim if group_offs is unbalanced but I think the align size set to 128 is to huge. Maybe we should align with the m dim padding align size of MXFP8 GEMM (https://github.com/AMD-AGI/Primus-Turbo/blob/8915c667840faf21e9c5345f08d36cdb9dfc3d84/primus_turbo/pytorch/kernels/gemm/gemm_fp8_impl.py#L298) ?

RuibinCheung (changes requested):

PR #333, opened 2026-05-09
fix: hipblaslt_gemm correct workspace size for MI300X

luiza-amd:
# Description

Issue arised during running Primux Flux1 FP8 local spec DDP training.

The HIPBLASLT FP8 GEMM backend on MI300X (gfx942) intermittently aborted training with HIPBLAS_STATUS_INVALID_VALUE from hipblasLtMatmul because Primus-Turbo passed a fixed 32 MiB scratch buffer to hipBLASLt, while hipBLASLt's own heuristic returned algorithms whose actual workspace requirement was up to ≈72 MiB for some FP8 mixed-precision shapes. The first reproducible victim was the (M=3072, N=3072, K=8192) E4M3 × E5M2 → BF16 grad-weight GEMM emitted by the FLUX-12B backward graph, which crashed within the first backward step of every run.

This PR resolves the crash by raising the gfx942 hipBLASLt workspace from 32 MiB → 128 MiB in get_hipblaslt_workspace_size_in_byte() and adds a targeted regression test that reproduces the exact failing problem shape, dtype mix, layout, granularity and backend so the issue cannot silently regress. gfx950 is left at 64 MiB and is unaffected.

# Fixes  (issue)

Fixes HIPBLAS_STATUS_INVALID_VALUE (code=3) from hipblaslt_gemm_impl (csrc/kernels/gemm/hipblaslt_gemm.hip:127) on MI300X / gfx942 when running FP8 mixed-precision GEMMs whose hipBLASLt-selected algorithm requires more than 32 MiB of workspace — observed on the FLUX-12B backward grad-weight GEMM (M=3072, N=3072, K=8192) E4M3 × E5M2 → BF16 (≈72 MiB required, hipBLASLt log: runContractionProblem: Input workspace size 33554432 is less than the required workspace size 75497472).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- csrc/kernels/gemm/hipblaslt_gemm.cu — bump the gfx942 / UNKNOWN return value of get_hipblaslt_workspace_size_in_byte() from 33554432 (32 MiB) to 134217728 (128 MiB), with an in-source comment explaining the FLUX-12B crash that motivates it. gfx950 stays at 64 MiB.

- tests/pytorch/ops/test_gemm_fp8.py — add test_gemm_fp8_hipblaslt_workspace_regression, a parameterized pytest that reuses _run_gemm_fp8_test to drive the exact failing production configuration (M=3072, N=3072, K=8192, layout NT, Format.HYBRID, torch.bfloat16, BackendType.HIPBLASLT, ScalingGranularity.TENSORWISE, auto_tune=False). The test fails with a RuntimeError from hipblaslt_gemm_impl if the 32 MiB cap is reintroduced and passes once the workspace is ≥ 75 497 472 bytes; it also validates SNR vs. the BF16 reference.

# Checklist:

### DIFF
--- a/csrc/kernels/gemm/hipblaslt_gemm.cu
+++ b/csrc/kernels/gemm/hipblaslt_gemm.cu
@@ -14,7 +14,16 @@ int64_t get_hipblaslt_workspace_size_in_byte() {
         return 67108864; // 64 MiB
     case GPUArch::GFX942:
     case GPUArch::UNKNOWN:
-        return 33554432; // 32 MiB
+        // 128 MiB. Some FP8 mixed-precision shapes on MI300X (gfx942) -- e.g.
+        // the (M=3072, N=3072, K=8192) E4M3 x E5M2 -> BF16 grad_weight GEMM
+        // emitted by the FLUX-12B backward graph -- cause hipBLASLt's
+        // heuristic to return an algorithm whose required workspace exceeds
+        // the previous 32 MiB cap (~72 MiB observed). hipblasLtMatmul then
+        // returns HIPBLAS_STATUS_INVALID_VALUE at execution time (logged by
+        // hipBLASLt as "Input workspace size ... is less than the required
+        // workspace size ..."). Bumping to 128 MiB covers that case with
+        // comfortable headroom for similar-sized algos.
+        return 134217728; // 128 MiB
     }
 }
 
--- a/tests/pytorch/ops/test_gemm_fp8.py
+++ b/tests/pytorch/ops/test_gemm_fp8.py
@@ -200,6 +200,49 @@ def _run_gemm_fp8_deterministic_test(
         GlobalBackendManager.reset()
 
 
+# Regression test for the hipBLASLt FP8 workspace shortage on MI300X (gfx942).
+#
+# Background:
+#   The HIPBLASLT FP8 backend uses a fixed-size workspace returned by
+#   ``get_hipblaslt_workspace_size_in_byte()`` in
+#   ``csrc/kernels/gemm/hipblaslt_gemm.cu``. Before commit
+#   ``fix/gemm-fp8-workspace-invalid-value-mi300x`` that value was 32 MiB
+#   on gfx942. For the specific FLUX-12B backward grad_weight GEMM
+#   ``(M=3072, N=3072, K=8192)`` with mixed E4M3 x E5M2 -> BF16
+#   tensorwise scaling, hipBLASLt's heuristic returns an algorithm whose
+#   actual workspace requirement is ~72 MiB. ``hipblasLtMatmul`` then
+#   returns ``HIPBLAS_STATUS_INVALID_VALUE`` at execution time (visible
+#   in hipBLASLt logs as ``runContractionProblem: Input workspace size
+#   33554432 is less than the required workspace size 75497472``).
+#
+#   This test reproduces the exact failing problem shape, dtype mix,
+#   layout, granularity and backend. It must succeed (no
+#   ``HIPBLAS_STATUS_INVALID_VALUE``, finite gradients) once the workspace
+#   is bumped to >= 75 497 472 bytes; it fails with a ``RuntimeError``
+#   from ``hipblaslt_gemm_impl`` if the regression is reintroduced.
+@pytest.mark.parametrize(
+    "m,n,k,layout,format,dtype,backend",
+    [
+        # FLUX-12B backward grad_weight GEMM that triggered the original crash.
+        (3072, 3072, 8192, "NT", Format.HYBRID, torch.bfloat16, BackendType.HIPBLASLT),
+    ],
+)
+def test_gemm_fp8_hipblaslt_workspace_regression(m, n, k, layout, format, dtype, backend):
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA not available")
+    _run_gemm_fp8_test(
+        m=m,
+        n=n,
+        k=k,
+        layout=layout,
+        format=format,
+        dtype=dtype,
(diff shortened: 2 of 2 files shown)


luiza-amd:
To motivate why workspace cap for 128MB both for MI300X and UNKNOWN architecture:

### 1. The failure mode is asymmetric

The two outcomes when you set the `MAX_WORKSPACE_BYTES` cap are not symmetric:

| Scenario | Outcome |
|---|---|
| Cap is **too small** for the algo hipBLASLt picks | Hard crash: `HIPBLAS_STATUS_INVALID_VALUE`, kills training |
| Cap is **larger than necessary** | hipBLASLt simply has more algorithms to choose from; it allocates only what its chosen algorithm actually needs |

So **the cost of being wrong on the low side is catastrophic** (entire 8-GPU training run dies); the cost of being wrong on the high side is essentially zero. In a function that has to pick one number for an unknown chip, you should bias toward the side where being wrong is cheap.

### 2. `MAX_WORKSPACE_BYTES` is a *cap*, not a forced allocation

This is the key point most people miss. We're calling:

```cpp
hipblasLtMatmulPreferenceSetAttribute(
    preference, HIPBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES,
    &workspace_size, sizeof(workspace_size));
```

That tells hipBLASLt: *"when picking an algorithm via the heuristic, you may consider any algo whose workspace is ≤ N bytes."* hipBLASLt then returns an algo whose actual workspace fits, and only that much scratch is consumed at `hipblasLtMatmul` time.

Concretely:
- For a small shape that needs 4 MiB of scratch, both `32 MiB` and `128 MiB` caps end up using **the same 4 MiB** at runtime.
- For a shape that needs 72 MiB, `32 MiB` cap **excludes that algo** (forcing a worse one — or in our crash case, hipBLASLt still returns the 72 MiB algo via heuristic and `hipblasLtMatmul` later blows up), and `128 MiB` cap **lets it pick freely** while still only using 72 MiB at runtime.

So bumping the `UNKNOWN` cap from 32 → 128 MiB is **free unless it's needed**, in which case it prevents a crash. There is no "wastes 96 MiB" downside.

### 3. What `UNKNOWN` actually means in practice

`get_current_arch()` returns `UNKNOWN` when it can't match the running chip against one of the explicitly enumerated values (`GFX942`, `GFX950`, etc.). In practice that path is hit for:

- **Newer or pre-release AMD Instinct GPUs** that ship before Primus-Turbo's enum is updated (e.g. successors to MI300X / MI355X). These chips will have similar or larger compute density, so their hipBLASLt algorithms are likely to require **similar or larger** workspaces — definitely not smaller.
- **Detection edge cases** (driver returns a string Primus-Turbo doesn't recognize, container runtime misreports arch, etc.) on machines that are *de facto* gfx942 — the most common deployment target today.

In both cases, the right answer is "behave at least as safely as gfx942". 128 MiB does that; 32 MiB does not.

### 4. The original code already treated `UNKNOWN` as "use the gfx942 default"

The pre-fix source had:

```cpp
case GPUArch::GFX942:
case GPUArch::UNKNOWN:
    return 33554432; // 32 MiB
```

The **fall-through grouping was deliberate** — the original author chose gfx942's value as the safe default for unknown archs, presumably for exactly the reasons above (gfx942 is the most-common production target). This PR doesn't change that policy; it only updates gfx942's number to one that empirically works. Splitting `UNKNOWN` off into its own bucket would be an *unrelated* behavior change, not motivated by any data, and would invent a new policy ("unknown archs are more conservative than known ones") that has no rationale.

### 5. The memory cost is negligible on every plausible target

| Target | 96 MiB extra cap as % of VRAM |
|---|---|
| MI300X (192 GB) | 0.047% |
| MI200 (128 GB) | 0.073% |
| Smallest plausible AMD compute GPU (16 GB) | 0.6% |

And again — per point 2 — this is the *cap*, not the actual allocation. The number above is the worst case if hipBLASLt happens to pick a 128 MiB-workspace algo. For most shapes the runtime allocation will be much smaller.

### 6. The alternatives are strictly worse or unmotivated

- **Keep `UNKNOWN` at 32 MiB**: Makes `UNKNOWN` strictly more crash-prone than any known arch. If the same FLUX-12B GEMM ever runs on a chip whose detection fails, it will still crash — defeating the purpose of having a fallback at all.
- **Set `UNKNOWN` to 64 MiB (match GFX950)**: Arbitrary. There's no evidence gfx950 is a better representative of unknown future archs than gfx942. Also still smaller than the empirically-required 72 MiB on the production GEMM that motivates this PR.
- **Make `UNKNOWN` an error (`PRIMUS_TURBO_CHECK(false, ...)`)**: Defensible long-term but breaks day-1 deployment on every new AMD chip until somebody updates the enum — a worse user experience than "works correctly with a slightly conservative cap".

### TL;DR

Bumping `UNKNOWN` to 128 MiB:

- **Costs nothing** in practice (it's a cap, not an allocation; max ~96 MiB extra cap = ≤0.05% of MI300X VRAM).
- **Prevents the same crash** on any unidentified chip whose hipBLASLt happens to behave like gfx942's — which is by far the most likely scenario, since `UNKNOWN` is overwhelmingly hit on gfx942-family deployments where detection didn't trigger.
- **Preserves the original code's design** of treating unknown archs as "use the gfx942 default", which is the conservative, safer option. Splitting them would invent a new policy with no data behind it.

So 128 MiB for both is the right choice; 32 MiB for `UNKNOWN` would just be a latent bug waiting to fire on the next chip generation.

xiaobochen-amd (approved):

xiaobochen-amd:
Thanks, LGTM.

PR #342, opened 2026-05-15
feat(gemm): Drop input_precision="ieee" from FP8 tl.dot

ChengYao-amd:
Drop `input_precision="ieee"` from FP8 `tl.dot`

In `_grouped_fp8_persistent_gemm_kernel`, the two `tl.dot` calls
inside the K-loop (one for the body, one for the K-tail) were
calling:

```python
acc += tl.dot(a, b, input_precision="ieee")
```

The `input_precision` knob in Triton selects between IEEE-compliant
fp32 reduction and TF32 reduction for **fp16 / bf16** inputs. For
FP8 (`E4M3` / `E5M2`) operands the kwarg has no defined behaviour
because FP8 already accumulates in fp32 unconditionally; on the AMD
Triton 3.7 backend it is at best a no-op and at worst forces a
defensive precision-cast in the MFMA selection path, blocking the
canonical `v_mfma_f32_*_f8f6f4` instruction.

The BF16 sibling kernel uses the documented knob
(`allow_tf32=ALLOW_TF32`) rather than `input_precision`, confirming
the FP8 kwarg was not chosen deliberately and is a stale copy.

This is a single-line change duplicated on both `tl.dot` sites and
does not affect numerical semantics: the accumulator is fp32 with
or without the kwarg, and FP8 inputs cannot satisfy any IEEE
precision constraint regardless of the flag.

### DIFF
--- a/csrc/include/primus_turbo/device/utils.cuh
+++ b/csrc/include/primus_turbo/device/utils.cuh
@@ -16,8 +16,9 @@ using namespace primus_turbo::dtype;
 // TODO: ASM
 template <typename T, const int N> PRIMUS_TURBO_DEVICE void load_data(const T *src, T *dst) {
     constexpr int BYTES = N * sizeof(T);
-    static_assert(BYTES == 1 || BYTES == 2 || BYTES == 4 || BYTES == 8 || BYTES == 16,
-                  "Only 1/2/4/8/16 bytes are supported.");
+    static_assert(BYTES == 1 || BYTES == 2 || BYTES == 4 || BYTES == 8 || BYTES == 16 ||
+                      BYTES == 32,
+                  "Only 1/2/4/8/16/32 bytes are supported.");
     if constexpr (BYTES == 1) {
         *reinterpret_cast<uint8 *>(dst) = *(reinterpret_cast<const uint8 *>(src));
     } else if constexpr (BYTES == 2) {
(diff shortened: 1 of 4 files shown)


csrc/include/primus_turbo/device/utils.cuh
```diff
@@ -16,8 +16,9 @@ using namespace primus_turbo::dtype;
 // TODO: ASM
 template <typename T, const int N> PRIMUS_TURBO_DEVICE void load_data(const T *src, T *dst) {
     constexpr int BYTES = N * sizeof(T);
-    static_assert(BYTES == 1 || BYTES == 2 || BYTES == 4 || BYTES == 8 || BYTES == 16,
-                  "Only 1/2/4/8/16 bytes are supported.");
+    static_assert(BYTES == 1 || BYTES == 2 || BYTES == 4 || BYTES == 8 || BYTES == 16 ||
+                      BYTES == 32,
+                  "Only 1/2/4/8/16/32 bytes are supported.");
```
xiaobochen-amd: My experience is that `BYTES=32` seems a bit suspicious. Wouldn't that make the memory accesses within a warp non-contiguous?

ChengYao-amd: I think coalescing is about cross-lane contiguity, not per-lane width. Benefit of bytes=32 is fewer loop iterations.

The 32 B per-thread load preserves coalescing because adjacent lanes still hit adjacent, aligned 32 B slots. The hardware decomposes the access into two contiguous dwordx4 issues, each of which is independently coalesced and which together cover the wavefront's 2048 B footprint exactly once. The change is a pure throughput optimization along the instruction-count and loop-overhead axis; it does not alter the cross-lane addressing pattern that coalescing depends on.

xiaobochen-amd (dismissed):

xiaobochen-amd (approved):

PR #345, opened 2026-05-26
refactor: moe_permute and DeepEPTokenDispatcher

zhenhuang12:
# Description

Refactor `moe_permute` and `DeepEPTokenDispatcher` to enable a fully nosync, CUDA-graph–capturable dispatch path.

Fixes # (issue)

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Code refactoring

## Changes

**`moe_permute` kernel**
- Count `num_dispatched_tokens` grid-wide via the existing decoupled-lookback prefix sum; last block writes with a plain store (no `atomicAdd`).
- Return `num_dispatched_tokens` as output instead of taking it as input — removes the host→device handshake.
- Add `probs_topk_stride` to support both `[T, num_local_experts]` multihot and `[T, num_topk]` topk-aligned probs layouts.
- Add `get_max_shmem_per_block` helper in `arch.h`.

**`_MoEPermute` op**
- Drop caller-supplied `num_dispatched_token_tensor`; the op now returns it.
- Infer `probs_topk_stride` from `probs.shape[1]`; raise on the ambiguous `num_local_experts == num_topk` case.
- Assert FP8 + probs-backward at the forward boundary.

**`DeepEPTokenDispatcher`**
- Fix: stop creating a per-rank singleton `dist.new_group` in the no-TP branch (leaked a collective per construction); reuse `ep_group`, set `tp_size = 1`.
- Rename `num_experts` → `tp_expanded_num_experts` to make TP-expansion explicit.
- Remove deprecated `permute_fusion`; rename `permute_max_token_num` → `num_permuted_tokens`; add `pad_multiple`.
- Lock `PRIMUS_TURBO_EP_FORCE_CURRENT_STREAM` after the first dispatcher; later conflicting constructors warn (C++ caches the env in a function-local `static`).
- Assert `probs` / `routing_map` / `token_indices` use the pre-TP-expansion expert id space — previously the reshape silently produced garbage.
- Document the nosync recipe: `deepep_num_worst_tokens > 0` + `num_permuted_tokens > 0` + `deepep_use_cuda_num_tokens_per_expert=True`.

**Tests**
- `test_moe_permute.py`: cover the new return signature and both probs layouts.
- `test_token_dispatcher.py`: cover the nosync path, no-TP group fix, and pre-TP shape asserts.

## Breaking changes

# Checklist


csrc/include/primus_turbo/arch.h
```diff
@@ -43,4 +43,11 @@ inline int32_t get_multi_processor_count(const int32_t device_id) {
     return num_cu;
 }
 
+inline int32_t get_max_shmem_per_block(const int32_t device_id) {
+    int32_t max_shmem = 0;
+    PRIMUS_TURBO_CHECK_HIP(
+        hipDeviceGetAttribute(&max_shmem, hipDeviceAttributeMaxSharedMemoryPerBlock, device_id));
```
RuibinCheung, line 49: The hipDeviceGetAttribute has some cpu overhead. If this func was called in every step please use static to avoid runtime overhead.

RuibinCheung (changes requested):

csrc/include/primus_turbo/arch.h
zhenhuang12: done

RuibinCheung (approved):
LGTM

PR #339, opened 2026-05-14
feat(gemm): optimize blockwise FP8 GEMM Triton kernels (gfx950 / MI355X)

ChengYao-amd:
# Optimize blockwise FP8 GEMM Triton kernels (gfx950 / MI355X)

PR branch: `dev/yaoc/optimize_fp8_blockwise_gemm_triton`
Base branch: `main` @ `06b8d3f` (`Add HYBRID FP8 format support for Triton backend in gemm and grouped_gemm`)
Commit on this branch: `opt: blockwise FP8 GEMM Triton kernels (gfx950 / MI355X)` (4 files, +597 / −46)

## 1. TL;DR

This PR optimizes the blockwise FP8 GEMM Triton path on gfx950 (MI355X).
Source-level changes touch four files only:

- `primus_turbo/triton/gemm/gemm_fp8_kernel.py`
- `primus_turbo/triton/quantization/quant_blockwise.py`
- `primus_turbo/pytorch/kernels/quantization/quantization_impl.py`
- `primus_turbo/pytorch/ops/gemm_fp8.py`

Aggregate impact on the 84-shape `bench_gemm_turbo.py` suite
(`PRIMUS_TURBO_GEMM_BACKEND=TRITON`, `--dtype fp8 --granularity blockwise`,
geomean over the 83 shapes both branches pass correctness on):

| Metric | `main` (gmean) | this PR (gmean) | delta |
|---|---:|---:|---:|
| Forward TFLOPS | 1470.6 | 1474.2 | **+0.25%** |
| Backward TFLOPS | 1018.7 | 1200.4 | **+17.83%** |
| Combined-step TFLOPS¹ | 1134.3 | 1278.4 | **+12.71%** |

¹ `Combined Step TFLOPS = 6·M·N·K / (fwd_ms + bwd_ms) / 1e9`.

Additional wins:

- **Correctness coverage**: `main` FAILs `TestID 47` (Llama-3.1-405B
  M=32768 N=106496 K=16384, dA=NaN). This PR PASSes all 84 shapes.
- **Per-shape distribution is uniformly positive on combined-step**:
  all 83 PASS-on-both shapes are ≥ +1% on combined-step
  (median +13.13%, p25 +10.95%, p75 +15.30%), 82 / 83 shapes are
  ≥ +5%. The backward delta is strictly positive on every shape
  (median +18.48%, min +3.22%, max +25.76%). Forward is centred near
  zero (median +0.63%, mean +0.28%) but 5 shapes carry a fwd
[...]

PR #353, opened 2026-06-02
feat: triton_grouped_gemm: add work-stealing variant with ws_mode API 

wenchenvincent:
# Description

  Adds a work-stealing (WS) variant of the persistent BF16/FP16 Triton grouped GEMM kernel
  (forward + variable-K backward) for AMD MI355X / MI350. Under all-gather × grouped-GEMM
  overlap, the static-stride persistent kernel was suffering a 1.7–2.0× slowdown vs isolated
  when RCCL channels contend for the same CUs; with WS enabled, fast CUs absorb work that the
   RCCL-throttled CUs would have done, recovering most of the slowdown at near-zero isolated
  overhead.

  The WS path is opt-in via work_steal=True on primus_turbo.ops.grouped_gemm, gated by a
  sync-free ws_mode heuristic so the right per-XCD / global / hierarchical claim policy is
  picked from tensor metadata without any GPU sync. Tests cover the four ws_mode values × 128
   shapes; isolated overhead is ~1% median on the 288-shape MoE sweep

co-authored with @scxiao

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

  - New WS forward kernel _grouped_bf16_persistent_gemm_kernel_ws in
  primus_turbo/triton/grouped_gemm/grouped_gemm_kernel.py — replaces the static block_id +=
  grid_size stride with tl.atomic_add tile claims (per-XCD slots + optional global tail),
  with the per-tile body refactored into a shared _process_grouped_gemm_tile helper so both
  static and WS reuse it.
  - New WS variable-K backward kernel _grouped_variable_k_gemm_kernel_ws — same WS scheme on
  the wgrad path; per-tile body extracted into _process_variable_k_tile.
  - Public API additions on grouped_gemm_triton_kernel /
  grouped_gemm_variable_k_triton_kernel: work_steal: bool, ws_mode: str ("auto" / "global" /
  "per-xcd" / "hierarchical"), ws_counter (caller-managed buffer for advanced multi-stream
  use), and num_cu (caps the persistent grid; was previously ignored).
  - Sync-free auto heuristic in primus_turbo/triton/grouped_gemm/ws_triton_heuristic.py —
  picks the WS mode from total_tiles/num_sms (derived from tensor metadata, no GPU reads).
  Thresholds tuned per kernel kind on the 288-shape MoE sweep.
  - Wrapper-managed singleton counter in
  primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py via _get_triton_ws_counter()
   — per-device int32 buffer sized (NUM_XCDS + 1) * COUNTER_STRIDE (padded to avoid XCD
  false-sharing); the kernel zeroes it before each launch.
  - work_steal / ws_mode plumbing through the autograd Function in
  primus_turbo/pytorch/ops/grouped_gemm.py so forward and backward both use WS when enabled.
  - waves_per_eu=0 on the WS launches to match the recent static-kernel optimization (PRs
  #339/#342). Without this the WS path was missing the static-side compiler-chosen occupancy
  and looked worse than it really was.
  - Defensive bound in _process_grouped_gemm_tile — if global_tile_id >= cumsum: return after
   the O(G) linear group scan, guarding against future callers that don't bound-check.
  - Tests in tests/pytorch/ops/test_grouped_gemm.py: 128 parametrized cases (B ∈ {4,8}, M ∈
  {1024,4096}, two (N,K) shapes, two balance settings, two trans_b settings, four ws_mode
  values) asserting bit-exact equality of WS forward + backward against the static reference,
   plus a G=1 smoke test.

# Checklist:

### DIFF
--- a/primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
+++ b/primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
@@ -25,6 +25,16 @@ from primus_turbo.triton.grouped_gemm.grouped_gemm_kernel import (
 
 _COMMON_SUPPORTED_DTYPES = (torch.float16, torch.bfloat16)
 
+# Supported schedule values exposed at the public op layer.
+#   "static"      -- static-stride persistent kernel (default).
+#   "work_steal"  -- work-stealing persistent kernel with the kernel-aware
+#                    heuristic that picks the WS sub-mode (per-XCD /
+#                    global / hierarchical) from tensor metadata.
+#                    Triton backend only in this build.
+_SUPPORTED_SCHEDULES: tuple[str, ...] = ("static", "work_steal")
+_TRITON_SUPPORTED_SCHEDULES: tuple[str, ...] = ("static", "work_steal")
+_NON_WS_SUPPORTED_SCHEDULES: tuple[str, ...] = ("static",)
+
 
 class GroupedGEMMCKBackend(KernelBackend):
     @staticmethod
@@ -36,12 +46,14 @@ class GroupedGEMMCKBackend(KernelBackend):
         trans_a: bool,
         trans_b: bool,
         num_cu: int | None,
+        schedule: str = "static",
         **kwargs,
     ) -> bool:
         supported = True
         supported &= a.dim() == 2 and b.dim() == 3
         supported &= a.dtype in _COMMON_SUPPORTED_DTYPES and b.dtype in _COMMON_SUPPORTED_DTYPES
         supported &= not trans_a
+        supported &= schedule in _NON_WS_SUPPORTED_SCHEDULES
         return supported
 
     @staticmethod
@@ -53,6 +65,7 @@ class GroupedGEMMCKBackend(KernelBackend):
         trans_a: bool,
         trans_b: bool,
         num_cu: int | None,
+        schedule: str = "static",
         **kwargs,
     ) -> torch.Tensor:
         return torch.ops.primus_turbo_cpp_extension.ck_grouped_gemm(
@@ -71,12 +84,14 @@ class GroupedGEMMVariableKCKBackend(KernelBackend):
         trans_b: bool,
         trans_c: bool,
         num_cu: int | None,
+        schedule: str = "static",
         **kwargs,
     ) -> bool:
         supported = True
         supported &= a.dim() == 2 and b.dim() == 2
         supported &= a.dtype in _COMMON_SUPPORTED_DTYPES and b.dtype in _COMMON_SUPPORTED_DTYPES
         supported &= trans_a and not trans_b
+        supported &= schedule in _NON_WS_SUPPORTED_SCHEDULES
         return supported
 
     @staticmethod
@@ -89,6 +104,7 @@ class GroupedGEMMVariableKCKBackend(KernelBackend):
         trans_b: bool,
         trans_c: bool,
         num_cu: int | None,
+        schedule: str = "static",
         **kwargs,
     ) -> torch.Tensor:
         if trans_c:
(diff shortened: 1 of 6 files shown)


xiaobochen-amd:
Hi @wenchenvincent ,

Thanks for the PR!

I have two points I'd like to discuss:

1. Turbo is generally very conservative about public API changes, especially when it comes to adding new arguments or modifying existing interfaces. We try to keep the API as simple and clean as possible. Is there a way to implement this feature using the existing API? Alternatively, we can discuss whether the interface can be simplified further.
2. How much performance improvement or benefit does this feature provide when enabled?

Thanks!

primus_turbo/pytorch/ops/grouped_gemm.py
```diff
@@ -29,6 +29,8 @@ def forward(
         group_offs: torch.Tensor,  # [B + 1,] int64
         trans_b: bool,
         num_cu: int | None,
+        work_steal: bool = False,
+        ws_mode: str = "auto",
```
xiaobochen-amd: Could we merge `work_steal` (bool) and `ws_mode` (str) into one `schedule: str = "static" | "auto"` (default `"static"`)? `global / per-xcd / hierarchical` are internal tuning modes — keep them on the low-level `grouped_gemm_triton_kernel` (`primus_turbo/triton/grouped_gemm/grouped_gemm_kernel.py`), not in the public API, so the heuristic can be re-tuned later without breaking the API.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
```diff
@@ -209,9 +237,23 @@ def execute(
...
         trans_b: bool,
         num_cu: int | None,
+        work_steal: bool = False,
+        ws_mode: str = "auto",
+        ws_counter: torch.Tensor | None = None,
         **kwargs,
     ) -> torch.Tensor:
-        return grouped_gemm_triton_kernel(a, b, group_offs, trans_b=trans_b)
+        if work_steal and ws_counter is None:
+            ws_counter = _get_triton_ws_counter(a.device)
```
xiaobochen-amd: The counter cache (`_triton_ws_counters` / `_get_triton_ws_counter`) is a kernel detail that shouldn't live in this layer — it even imports `COUNTER_STRIDE`/`NUM_XCDS` from the triton module. Could we move it into `primus_turbo/triton/grouped_gemm/grouped_gemm_kernel.py` and just pass `schedule` through here?

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
xiaobochen-amd, line 220: Non-Triton backends currently swallow the WS args via `**kwargs`, so `schedule="auto"` on CK/hipblaslt quietly falls back to static. Could `can_handle` spell out which schedules each backend supports?
- Triton: `supported &= schedule in ("static", "auto")`
- CK/hipblaslt: `supported &= schedule == "static"`

Then if a user explicitly selects a non-Triton backend together with `auto`, dispatch raises at backend selection (via the existing mechanism) instead of silently running static. Also, pass `schedule` as an explicit arg to `can_handle`/`execute` rather than `kwargs.get`.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
wenchenvincent: Done.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_impl.py
wenchenvincent: Done.

primus_turbo/pytorch/ops/grouped_gemm.py
wenchenvincent: Updated. Used "work_steal" instead of "auto" as it is clearer.

xiaobochen-amd:
LGTM！ Thanks

xiaobochen-amd (approved):

PR #361, opened 2026-06-06
opt(build): enable ninja incremental builds by disabling force-rebuild

jasainio:
PyTorch's `BuildExtension.finalize_options` sets `self.force = True` when ninja is used, causing setuptools to unconditionally rebuild all extensions on every install. Override with `self.force = False` in `TurboBuildExt` so ninja's dependency tracking (`.d` files) can skip unchanged translation units.

- Single-file rebuilds drop from ~5 min to ~30 seconds in the Primus-LM dev container
- No behavioural change for clean builds (ninja still compiles everything)
- Inline comment in `tools/build_ext.py` documents the override rationale

# Description

PyTorch's `BuildExtension.finalize_options()` always sets `self.force = True` for ninja-backed extensions. setuptools then re-runs every translation unit through `hipcc` on every `pip install --no-deps -e .`, even when nothing changed. `TurboBuildExt` now resets `self.force = False` after calling `super().finalize_options()`, letting ninja's `.d` dependency tracking skip up-to-date objects.

Verified in the Primus-LM dev container: a single-file edit + `pip install --no-deps --no-build-isolation -e .` now completes in ~30 s instead of ~5 min. Verified end-to-end as part of `dev/mxfp4-aiter-preshuffle-min` @ 026e136 (Flux 12B MXFP4 1000-step exit 0, FP8 delayed 1000-step exit 0).

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- `tools/build_ext.py`: in `TurboBuildExt.finalize_options`, set `self.force = False` after `super().finalize_options()` so ninja's incremental dependency tracking is honoured

# Checklist:

- [ ] I have made corresponding changes to the documentation (N/A - no user-facing API change; build behaviour only)
- [ ] I have added tests that prove my fix is effective or that my feature works (N/A - 1-line build flag; verified manually by timing rebuilds in the dev container)

### DIFF
--- a/tools/build_ext.py
+++ b/tools/build_ext.py
@@ -81,8 +81,6 @@ class BuildExtension(build_ext):
 
     def finalize_options(self) -> None:
         super().finalize_options()
-        if self.use_ninja:
-            self.force = True
 
     def build_extensions(self) -> None:
 
@@ -562,6 +560,7 @@ class TurboBuildExt(BaseBuildExtension):
 
     def finalize_options(self):
         super().finalize_options()
+        self.force = False
         if self.inplace:
             self.build_temp = str(self.PROJECT_ROOT / "build" / "temp")
             self.build_lib = str(self.PROJECT_ROOT / "build" / "lib")


RuibinCheung:
@xiaobochen-amd Please help me take a look.

xiaobochen-amd:
Thanks for the incremental-build optimization — I get the direction, but I have a concern I'd like to align on first.

One small note first: the change to the local BuildExtension.finalize_options (removing self.force = True) is effectively dead code in our Torch build path. When Torch is available, TurboBuildExt inherits from torch.utils.cpp_extension.BuildExtension (via _select_base_build_ext), so the local BuildExtension isn't in the MRO at all. The only line that actually takes effect is self.force = False in TurboBuildExt.finalize_options, which overrides Torch's setting.

And that's where my concern is. PyTorch intentionally sets force = True for ninja builds ([code here](https://github.com/pytorch/pytorch/blob/main/torch/utils/cpp_extension.py), BuildExtension.finalize_options) so that ninja is the single source of truth for incrementality, deliberately bypassing setuptools' own coarser check. Setting force = False brings that check back, and since our extensions only pass sources= (no depends=), setuptools decides solely from source-file mtimes. Two cases then get silently skipped, with no error:

- Editing only a header (no source touched) → not rebuilt.
- Changing build flags (GPU_ARCHS, DISABLE_ROCSHMEM, etc.) without touching sources → stale binary with old flags.

For a kernel-heavy project like ours, this "silently get a stale binary" risk worries me. I'd rather stay conservative on stability and keep aligned with PyTorch here — can we discuss this approach a bit more?

jasainio:
You're right -- I missed that the local `BuildExtension` path is dead code and only `TurboBuildExt.finalize_options` matters. The stale-binary risk from header/flag changes is a real concern for a kernel-heavy project, especially since our `setup.py` passes no `depends=` to `HIPExtension` and ROCm hipcc doesn't emit depfiles.

Closing this in favor of PR #369, which takes the safer approach (ccache/sccache as a compiler launcher -- composes with `force=True` without bypassing correctness). Thanks for the thorough review.

PR #322, opened 2026-04-30
feat(quantization): add fused amax+scale FP8 tensorwise quantize

jasainio:
Fuse the abs-max reduction and scale computation into a single final-round kernel, reducing total kernel launches from 3 to 2 for the tensorwise FP8 quantization path.

- Add `reduce_amax_final_scale_kernel` that computes `scale = fp8_max / amax` and `scale_inv = amax / fp8_max` directly in the reduction epilogue
- Add `reduce_amax_and_compute_scale` dispatcher with multi-round support for large tensors (reuses existing `reduce_row_kernel` for intermediate rounds)
- Register `quantize_fp8_tensorwise_fused` op in PyTorch bindings (CUDA + Meta)
- Expose `quantize_fp8_fused()` public Python API with fallback for non-tensorwise granularities
- Add tests for fused path: tensorwise + amax correctness (66 cases)
- Add benchmark script comparing original 3-kernel vs fused 2-kernel latency

# Description

Fuse the abs-max reduction and scale computation into a single HIP kernel for the tensorwise FP8 quantization path. The original path required 3 kernel launches (`reduce_row` -> `compute_scale_from_amax` -> `quantize_tensorwise_impl`). The fused path eliminates the standalone scale computation by folding it into the final reduction round, reducing the total to 2 kernel launches (`reduce_amax_and_compute_scale` -> `quantize_tensorwise_impl`). For large tensors that require multi-round reduction, intermediate rounds reuse the existing `reduce_row_kernel`, and only the final round uses the fused kernel. A workspace size debug assertion and ping/pong buffer aliasing safety are documented in the kernel code.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Add `reduce_amax_final_scale_kernel` in `csrc/kernels/quantization/quantization.cu` that fuses abs-max reduction with scale/scale\_inv computation in the epilogue
- Add `reduce_amax_and_compute_scale` host dispatcher with single-tile fast path and multi-round support using ping/pong workspace buffers
- Add workspace size debug assertion and ping/pong aliasing safety comment for multi-round reduction
[...]

PR #364, opened 2026-06-06
feat(quantize): add fused FP8 quantization kernels with amax+scale and cast+transpose

jasainio:
Add fused FP8 cast+transpose+amax kernels enabling Primus's delayed-scaling float8 local-spec path. Tiled shared-memory C++/HIP backend + autotuned Triton backend, with optional amax output for delayed scaling. Includes meta tensor support for `torch.compile`.

- Adds `cast_transpose_fp8` Triton kernel (registered as `@triton_op`, Inductor-transparent)
- Adds tiled shared-memory C++/HIP backend `quantization_cast_transpose.cu`
- Extends `quantize_fp8_tensorwise_fused` with optional amax output for delayed scaling
- Public Python API in `primus_turbo/pytorch/ops/quantization.py`
- Tests + benchmarks for both backends
- Builds on #322 (`opt/quantize/fused-amax-scale`)

> Depends on #322 (`opt/quantize/fused-amax-scale`). Merge #322 first or rebase this PR after it lands.

# Description

Primus's float8 local-spec path (`primus.backends.megatron.core.extensions.primus_turbo_float8_local`) needs a fused FP8 `cast + transpose + amax` op to avoid bf16 round-trips between the forward and column-major-transpose paths. This PR adds two backends:

1. **Triton backend** (`primus_turbo/pytorch/kernels/quantization/cast_transpose_fp8.py`) -- registered as `@triton_op("primus_turbo::cast_transpose_fp8_triton")` so it is Inductor-transparent. Autotuned tiling for production Flux 12B shapes. Optional amax output for delayed scaling.
2. **C++/HIP backend** (`csrc/kernels/quantization/quantization_cast_transpose.cu`) -- tiled shared-memory implementation as a reference + fallback.

Also extends `quantize_fp8_tensorwise_fused` with an optional `amax_out` output so delayed-scaling pipelines can update their amax history in the same kernel.

Verified end-to-end as part of `dev/mxfp4-aiter-preshuffle-min` @ 026e136 (Flux 12B FP8 delayed 1000-step exit 0; Primus float8 local-spec path imports `cast_transpose_fp8_triton`).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add `csrc/kernels/quantization/quantization_cast_transpose.cu` (tiled shared-memory `cast_transpose_fp8` HIP kernel)
- Extend `csrc/kernels/quantization/quantization_tensorwise.cu` with `quantize_fp8_tensorwise_fused` amax-output variant (vectorized with UNROLL pack_size 8/4/2/1 + tail loop)
- Register C++ + Meta bindings (`csrc/pytorch/quantization/quantization.cpp`, `..._meta.cpp`); update `csrc/pytorch/extensions.h` and `csrc/pytorch/bindings_pytorch.cpp`
- Add `primus_turbo/pytorch/kernels/quantization/cast_transpose_fp8.py` (Triton backend, autotuned, registered as `@triton_op`)
- Expose public Python APIs in `primus_turbo/pytorch/ops/quantization.py`
- Add `tests/pytorch/ops/test_quantization.py` and `tests/pytorch/ops/test_cast_transpose_fp8.py` coverage
- Add `benchmark/ops/bench_cast_transpose_fp8.py`; extend `bench_fp8_quantize_tensorwise.py`

# Checklist:

- [ ] I have made corresponding changes to the documentation (N/A - new Python API has docstrings; consumed via Primus float8 local spec)


xiaobochen-amd:
@jasainio  Thanks for the PR. Before getting into implementation details, two higher-level things I'd like to align on first.

1. **quantize_fp8_fused**. For tensorwise, the scale is just a single scalar, so this fusion mainly saves the launch of the scale kernel — the memory traffic is basically unchanged (still one full amax pass + one full quantize pass). For a memory-bound op the gain should be pretty limited, and it's not really new functionality. So I'd rather not add this as a separate public API for now.

2. **cast_transpose_fp8** — could you clarify what it's for and where it belongs:

- What problem does it actually solve?
- Is it the same kind of thing as the existing **quantize_fp8_with_trans**? If so, can it just go in as the TENSORWISE branch of quantize_fp8_with_trans (dispatched by granularity), instead of adding a new API with a different style? We're pretty conservative about growing the public API surface and would rather not introduce a second naming/signature style in ops/.
- If it gives a speedup, could you share some e2e numbers?

cc: @RuibinCheung

jasainio:
Thanks for the feedback @xiaobochen-amd.

**1. `quantize_fp8_fused`** -- Agreed, dropped entirely. As you noted, tensorwise scale is a single scalar so the fusion only saves the scale kernel launch (~2 us). Not worth a new API.

**2. `cast_transpose_fp8` -- what it solves and where it belongs:**

**(a) Problem it solves:** In delayed-scaling FP8 training, every GEMM input needs three things: FP8-quantized tensor, its contiguous transpose (for the backward GEMM), and the current abs-max (to compute next iteration's scale). Without fusion, this is 3+ separate kernels reading the full tensor each time. The fused Triton kernel (`cast_transpose_fp8_triton`) does all three in a single pass -- one HBM read of the input produces `(fp8_out, fp8_transposed, amax)`.

We have a delayed-scaling FP8 linear module that calls this kernel for every GEMM input in both forward and backward -- quantizing weight, input, and grad_output each with fused transpose + amax. That training integration (in Primus-LM) depends on these primitives landing in Primus-Turbo first.

**(b) Integration as TENSORWISE branch:** Done. I've added it as `quantize_fp8_with_trans(..., granularity=ScalingGranularity.TENSORWISE, scale=..., amax_out=...)`. No new public API name -- it dispatches internally to the Triton kernel.

**(c) Performance numbers (MI355X):**

| Shape | Unfused (us) | Triton Fused (us) | Speedup |
|-------|-------------|-------------------|---------|
| 4096x3072 | 143.6 | 47.6 | 3.0x |
| 4096x12288 | 451.1 | 51.7 | 8.7x |
| 12288x4096 | 455.2 | 51.8 | 8.8x |
| 16384x3072 | 450.4 | 52.1 | 8.6x |
| 8192x8192 | 564.2 | 65.0 | 8.7x |
| 32768x3072 | 868.9 | 102.2 | 8.5x |
| 3072x12288 | 352.5 | 45.8 | 7.7x |

The "unfused" baseline is `quantize_fp8_tensorwise(x, dtype, None)` + `.t().contiguous()` + amax reduction. In production the scale is pre-computed (delayed scaling), so the fair comparison is cast + transpose + amax = 3 kernels / 2-3 HBM reads. Still a clear win for the single fused kernel.

End-to-end with a downstream GEMM (quantize+transpose then `_scaled_mm`):

| Shape | Unfused (us) | Triton (us) | Saved |
|-------|-------------|-------------|-------|
| 16384x3072 | 578 | 163 | -415 us |
| 16384x12288 | 3476 | 1855 | -1621 us |
| 32768x3072 | 1122 | 334 | -789 us |

**Reproduce:**
```
python benchmark/ops/bench_cast_transpose_fp8.py --warmup 30 --repeat 300 --gemm
```

**Other changes in this push:**
- Removed `bias_gelu_cast_transpose_fp8` and `dbias_dgelu_cast_transpose_fp8` from this PR (will submit as follow-up)
- Removed `cast_transpose_amax` backward-compat alias
- Updated `__all__` (removed `quantize_fp8_fused` and `cast_transpose_fp8`)
- All correctness checks pass (7/7 shapes)

RuibinCheung:
May I know why you need add this op? 
The tensorwise gemm in turbo not likes TE that only support TN layout. It supports TN, NN and NT layout so we don't need do extra transpose in tensorwise granularity. So I think the cast+transpose op is not necessary.

jasainio:
You're right that hipBLASLt supports TN, NN, and NT layouts for tensorwise FP8. However, the cast+transpose is needed for **performance**, not correctness.

In FP8 training, the forward GEMM naturally uses NT layout (`C = A @ B^T`), but the two backward GEMMs use different layouts: NN for grad_input (`dX = dY @ W`) and TN for grad_weight (`dW = X^T @ dY`). When we profiled an FP8 training workload on MI355X, the backward GEMMs running in their natural NN/TN layouts were **~2x slower** than when all GEMMs are normalized to NT via pre-transposition:

| op | natural layout | all-NT TFLOPS/s | natural layout TFLOPS/s | ratio |
|---|---|---:|---:|---:|
| forward | NT (F,T,F) | 2739 | 2817 | 1.03x |
| grad_input | NN (F,F,F) | 2872 | 1569 | 0.55x |
| grad_weight | TN (T,F,T) | 2481 | 1306 | 0.53x |

(Averaged across 8 GEMM shape classes on MI355X. Forward uses NT in both configurations, confirming parity.)

NT in row-major maps to TN in hipBLASLt's column-major convention, which selects the fastest Tensile kernel on MI355X. The other layouts are functional but the heuristic picks slower kernel tiles.

This is consistent with what we see elsewhere in Primus-Turbo: `FP8GemmRowFunction.backward`, the FP4 GEMM backends, and the CK blockwise backend all normalize to NT layout. This kernel provides the fused quantize + transpose + amax primitive that downstream FP8 linear layers need to normalize backward GEMMs to NT without an additional memory pass. For delayed scaling, the fused amax is also needed to compute next iteration's scale without a separate reduction kernel.

Happy to discuss further or run a controlled layout benchmark directly on `gemm_fp8_impl`.

jasainio:
Closing this PR — we're not landing this kernel set on `main`.

The fused FP8 cast / cast+transpose kernels here were only ever needed by a specific downstream delayed-scaling integration, not by Primus-Turbo's core API. Rather than carry a public-API surface for kernels that no general Turbo consumer uses, we've moved these self-contained Triton kernels to the one consumer that needs them. That keeps Turbo `main` focused on the broadly-used quantization ops and avoids maintaining an integration-only path here.

Nothing on `main` depended on this branch, so there's no follow-up cleanup required. Thanks to everyone who reviewed!

PR #335, opened 2026-05-09
feat: add quantized tensor support

RuibinCheung:
# Description

This PR introduces a unified `QuantizedTensor` wrapper class as the canonical container for all FP8 / FP4 low-precision paths. As part of the work, the FP8/FP4 quantization kernels are reorganized, a new rowwise dequantize C++ binding and an optimized col-major dequant kernel are added, and all GEMM / grouped GEMM / MoE permutation ops are migrated to the new `QuantizedTensor` API.

Motivation:

- The previous `Float8Tensor` was still under development, did not cover FP4 or the full set of granularities (rowwise / tensorwise / MX-blockwise), and could not be passed as a first-class citizen through grouped GEMM / MoE permutation paths.
- On the C++ side, `quantization.cu` lumped tensorwise / rowwise / MXFP8 / MXFP4 into a single TU, leaving shared `QuantOp` / `amax→scale` helpers tightly coupled and hard to reuse or extend.
- The col-major path of rowwise dequantize processed only one row per block, so re-loading `scale_inv` consumed a non-trivial fraction of bandwidth on tall-M shapes.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Infra/Build change
- [x] Code refactoring

> Breaking change notes: the legacy `primus_turbo.pytorch.core.float8_tensor.Float8Tensor` has been removed; downstream callers must migrate to the new `QuantizedTensor` API. The C++ schemas of `quantize_mxfp4*` / `quantize_mxfp8*` have a new required argument `padding_align_size`.

### New unified `QuantizedTensor` wrapper

- Add `primus_turbo/pytorch/core/quantized_tensor.py`:
  - `QuantizedTensor`: a wrapper subclass built on `torch.Tensor._make_wrapper_subclass`, carrying `_data` / `_scale_inv` together with metadata such as `orig_dtype` / `dest_dtype` / `granularity` / `block_size` / `scaling_recipe` / `quantized_axis`, plus `group_lens` / `group_offs` for the grouped case.
[...]

PR #366, opened 2026-06-06
opt(gemm): add AITER MXFP4 preshuffle fast path

jasainio:
Reintroduces the GEMM-side preshuffle fast path that PR #335 dropped when consolidating FP4 GEMM dispatch. With AITER pinned and autotune disabled, `FP4GemmMXFunction` now emits already-shuffled scales / B-data from the fused quantize kernel and `gemm_fp4_impl(preshuffled=True)` calls `aiter.gemm_a4w4(bpreshuffle=True)` directly, skipping 3 per-GEMM shuffle kernel launches.

- ~10-15% per-GEMM speedup on Flux 12B shapes (microbench)
- Recovers the ~5% step-time regression reported on `flux_ddp_mxfp4_full_rebase_verify_263`
- End-to-end verified on Flux 12B MXFP4 1000-step (224 BpreShuffle kernel binds, exit 0) under `dev/mxfp4-aiter-preshuffle-min` @ 026e136
- HipBLASLt backend `can_handle` returns `False` on `preshuffled=True` so Path 1 raises cleanly instead of silently corrupting
- 4 commits: fast path + 3 supporting fixes (empty backend env var; `_clear_env_cache` re-export; FP4 GEMM `torch.compile` backward compat)

# Description

PR #335 consolidated FP4 GEMM dispatch but removed the preshuffle fast path, causing a ~5% step-time regression on Flux 12B MXFP4. This PR re-adds the fast path under a clean gate that activates only when (a) AITER is the pinned backend and (b) autotune is disabled -- the exact configuration where preshuffle is safe and beneficial.

When `enable_preshuffle()` returns `True`, the fused quantize kernel produces already-shuffled scales / B-data, and `gemm_fp4_impl(preshuffled=True)` dispatches directly to `aiter.gemm_a4w4(bpreshuffle=True)`. This skips 3 per-GEMM shuffle launches. HipBLASLt's `can_handle` returns `False` on `preshuffled=True` (the C++ extension has no preshuffle support), so any accidental routing to Path 1 raises a clean error instead of silently producing garbage.

The 3 supporting commits address: handling of empty `PRIMUS_TURBO_GEMM_BACKEND` env var, re-export of `_clear_env_cache` after the lru_cache removal, and FP4 GEMM backward compatibility under `torch.compile`.

Verified end-to-end as part of `dev/mxfp4-aiter-preshuffle-min` @ 026e136: Flux 12B MXFP4 1000-step exit 0 with 224 `BpreShuffle` AITER kernel binds confirmed.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring (performance optimization: re-introduces a fast path that PR #335 dropped)

## Changes

- `primus_turbo/pytorch/kernels/gemm/gemm_fp4_impl.py`: add `enable_preshuffle()` helper (gates on AITER pin + autotune-off); thread `preshuffled: bool = False` through `custom_op`, `register_fake`, dispatcher kwargs, `make_key` (cache invalidation). AITER backend branches on `preshuffled` True/False; HipBLASLt backend `can_handle` returns `False` on `preshuffled=True`
- `primus_turbo/pytorch/ops/gemm_fp4.py`: wire `enable_preshuffle()` into the 6 `ScalingRecipe` constructors (only `shuffle_scale` + `shuffle_out` flip) and into the 3 `gemm_fp4_impl(...)` calls. `default_backend` switches to AITER under preshuffle. When deriving the col-wise transpose tensor and `shuffle_scale=True` is active, prefer the original BF16 input over `a_fp4.dequantize()` (which would consume the shuffled scale as canonical)
- `primus_turbo/pytorch/core/backend.py`: treat empty `PRIMUS_TURBO_GEMM_BACKEND` env var as missing; expose `_clear_env_cache` after the `lru_cache` removal
- `primus_turbo/pytorch/kernels/quantization/quantization_impl.py`: minor wiring for shuffled-output path
- Add 4 new tests in `tests/pytorch/ops/test_gemm_fp4.py`: `test_gemm_fp4_impl_aiter_preshuffle_parity` (bit-for-bit parity True vs False on Flux shapes), `test_gemm_fp4_impl_hipblaslt_preshuffle_opt_out` (locks HipBLASLt rejection contract), `test_gemm_fp4_mx_blockwise_quantized_tensor_aiter_preshuffled` (QuantizedTensorPair contract), and end-to-end wiring assertion in existing `test_gemm_fp4_mx_blockwise` matrix
- `tests/pytorch/core/test_global_backend_manager.py`: env-cache assertions
- `benchmark/ops/bench_gemm_fp4_preshuffle.py`: microbenchmark for cold/warm-cache preshuffle vs non-preshuffle

# Checklist:

- [x] I have made corresponding changes to the documentation (consumer docs updated in primus-tiger-training-internal `dev_workflow.md`)
- [x] New and existing unit tests pass locally with my changes; end-to-end verified on Flux 12B MXFP4 1000-step (exit 0, 224 BpreShuffle kernel binds)


primus_turbo/pytorch/core/backend.py
```diff
@@ -122,8 +124,20 @@ def _extract_backend_from_env(cls, env_value: str) -> Dict[PrecisionType, Backen
             for value in _PRECISION_TYPE_MAPPING.values():
                 precision_backend_dict[value] = BackendType[env_value.upper()]
 
+        cls._env_cache[env_value] = precision_backend_dict
         return precision_backend_dict
 
+    @classmethod
+    def _clear_env_cache(cls) -> None:
```
RuibinCheung, line 131: Why need add this func ?

primus_turbo/pytorch/kernels/gemm/gemm_fp4_impl.py
```diff
@@ -185,9 +207,39 @@ class GEMMFP4KernelDispatcher(AutoKernelDispatcher):
...
     @classmethod
-    def make_key(cls, a, b, trans_a, trans_b, trans_c, out_dtype, granularity, **kwargs):
+    def make_key(cls, a, b, trans_a, trans_b, trans_c, out_dtype, granularity, preshuffled=False, **kwargs):
         m, n, k = get_gemm_logical_shape(a, b, trans_a, trans_b)
-        return (m, n, k, a.dtype, b.dtype, out_dtype, trans_a, trans_b, trans_c, granularity)
+        return (m, n, k, a.dtype, b.dtype, out_dtype, trans_a, trans_b, trans_c, granularity, preshuffled)
+
+
+def enable_preshuffle() -> bool:
+    """Return True iff the AITER FP4 preshuffle fast path is safe to take.
```
RuibinCheung: Could you try to simplify this comment ?

benchmark/ops/bench_gemm_fp4_preshuffle.py
```diff
@@ -0,0 +1,315 @@
+###############################################################################
```
RuibinCheung: We want to keep the benchmark/ops directory simple. Could you removed this file?

RuibinCheung (changes requested):

primus_turbo/pytorch/core/backend.py
jasainio: `_clear_env_cache()` replaces `@functools.lru_cache` which was removed because `lru_cache` on a classmethod is opaque to `torch._dynamo` and causes a graph break when `get_gemm_backend()` is called inside a `torch.compile` region. The call chain is: `enable_preshuffle()` → `GlobalBackendManager.get_gemm_backend()` → `_extract_backend_from_env()` — with `lru_cache` wrapping that last call, dynamo cannot trace through it and inserts a graph break.

The replacement is a plain `dict` class variable (`_env_cache`) that dynamo can trace through. The `_clear_env_cache()` method exposes the invalidation API so tests can reset state between parametrize runs (previously they called `_extract_backend_from_env.cache_clear()`).

If you prefer, I can inline the `.clear()` call directly into the test fixtures and drop the method.

primus_turbo/pytorch/kernels/gemm/gemm_fp4_impl.py
jasainio: Done — condensed to 5 lines in 291a7b0.

benchmark/ops/bench_gemm_fp4_preshuffle.py
jasainio: Done — removed in 291a7b0.

wenchenvincent:
@jasainio  Does this PR target training or inference? For training, the data will need to be preshuffled every iteration.

jasainio:
This targets **training**. The preshuffle is not a one-time preprocessing step -- it's fused into the quantize kernel that already runs every forward and backward call.

The existing `quantize_mxfp4` C++ kernel already accepts `shuffle_scale` / `shuffle_out` flags (added as part of the AITER MXFP4 integration before this PR). What this PR adds is the **wiring** that activates those flags:

- `FP4GemmMXFunction.forward` passes `shuffle_scale=True` / `shuffle_out=True` in the `ScalingRecipe` to `QuantizedTensor.quantize()` every iteration ([gemm_fp4.py L59-101](https://github.com/AMD-AGI/Primus-Turbo/blob/opt/gemm/fp4-aiter-preshuffle-fastpath/primus_turbo/pytorch/ops/gemm_fp4.py#L59-L101))
- The quantize kernel produces already-shuffled FP4 data + scales in a single fused pass
- `gemm_fp4_impl(preshuffled=True)` then calls `aiter.gemm_a4w4(bpreshuffle=True)` directly, skipping 3 standalone shuffle kernel launches that would otherwise run per-GEMM

The backward pass does the same: `grad_out` is re-quantized with `shuffle_scale=preshuffle` every iteration ([gemm_fp4.py L189-250](https://github.com/AMD-AGI/Primus-Turbo/blob/opt/gemm/fp4-aiter-preshuffle-fastpath/primus_turbo/pytorch/ops/gemm_fp4.py#L189-L250)).

So the cost model is: zero extra work (shuffle is folded into the quantize kernel we were already paying for), minus 3 kernel launches per GEMM.

primus_turbo/pytorch/core/backend.py
RuibinCheung: ACK

tests/pytorch/ops/test_gemm_fp4.py
```diff
@@ -266,3 +298,418 @@ def test_gemm_fp4_mx_blockwise_quantized_tensor(m, n, k, layout, format, dtype,
...
+        # Both paths converge on the same aiter.gemm_a4w4(bpreshuffle=True)
+        # consuming the same byte-identical shuffled inputs -> outputs match.
+        assert torch.equal(out_no_pre, out_pre), (
+            "preshuffled=True must match preshuffled=False bit-for-bit " "on the AITER backend"
+        )
+    finally:
+        GlobalBackendManager.reset()
+
+
+def test_gemm_fp4_impl_hipblaslt_preshuffle_opt_out():
```
RuibinCheung: Please remove this unit test case.

primus_turbo/pytorch/kernels/quantization/quantization_impl.py
```diff
@@ -533,6 +533,9 @@ def quantize_mxfp4_impl(
     else:
         assert axis is None, "The axis must be None when with_trans is True."
 
+    if not x.is_contiguous():
```
RuibinCheung: Dose x will be a non-contiguos tensor ? The contiguous may cause silent performance regression. If we need check x is contiguous please use assert `x.is_contiguous()`.

RuibinCheung (changes requested):

tests/pytorch/ops/test_gemm_fp4.py
jasainio: Removed in bf138d3. The production guard in `GEMMFP4HipBLASLtBackend.supports()` that rejects `preshuffled=True` is untouched -- only the test is removed.

primus_turbo/pytorch/kernels/quantization/quantization_impl.py
jasainio: Good catch -- changed to `assert x.is_contiguous(), "The x tensor must be contiguous."` in bf138d3 to match the convention in the rest of the file (lines 51, 68, 454).

On `main`, `quantize_mxfp4_impl` has no contiguity check at all and the C++ kernel would silently produce wrong results on non-contiguous input. The assert makes the requirement explicit. The forward callers in `FP4GemmMXFunction` don't force contiguity on user inputs, but in practice Megatron/Primus-LM always feeds contiguous activations. If a non-contiguous case turns up, the fix would be adding `.contiguous()` at the caller level rather than hiding it inside the kernel.

RuibinCheung (dismissed):
LGTM if CI is green.

RuibinCheung (approved):
LGTM

PR #369, opened 2026-06-08
feat(build): add optional ccache/sccache compiler launcher support

jasainio:
# Description

Add optional ccache/sccache compiler launcher support to speed up incremental Primus-Turbo rebuilds during development. When `PRIMUS_TURBO_COMPILER_LAUNCHER` is set, the specified cache tool is prepended to both `cxx` and `nvcc` compiler commands in the generated Ninja build file. Default behavior is unchanged when the variable is unset.

## Type of change

- [x] Infra/Build change

## Changes

- Add `_get_compiler_launcher()` helper in `tools/build_ext.py` that reads `PRIMUS_TURBO_COMPILER_LAUNCHER` env var
- Modify `_write_ninja_file()` to prepend the launcher to `cxx` and `nvcc` Ninja variables when set
- Document `PRIMUS_TURBO_COMPILER_LAUNCHER` env var in README.md Development section

### Usage

```bash
PRIMUS_TURBO_COMPILER_LAUNCHER=ccache pip install --no-build-isolation -e .
```

### Verification

Tested in `dev_primus_ubuntu_263` container (rocm/primus:v26.3, MI355X):

| Scenario | Result |
|---|---|
| First build with ccache | 55 cacheable calls, 55 misses (expected) |
| Clean rebuild with ccache | 55 cacheable calls, all cache hits |
| Build without env var | `cxx = /opt/rocm-7.2.1/bin/hipcc` (no prefix, default preserved) |

# Checklist:

- [x] I have made corresponding changes to the documentation

### DIFF
--- a/tools/build_ext.py
+++ b/tools/build_ext.py
@@ -54,6 +54,14 @@ def get_cxx_compiler():
     return compiler
 
 
+def _get_compiler_launcher():
+    launcher = os.environ.get("PRIMUS_TURBO_COMPILER_LAUNCHER")
+    if launcher:
+        print(f"[Primus-Turbo] Using compiler cache: {launcher}")
+        return launcher
+    return None
+
+
 class BuildExtension(build_ext):
     @classmethod
     def with_options(cls, **options):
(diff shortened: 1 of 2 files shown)


tools/build_ext.py
```diff
@@ -54,6 +54,14 @@ def get_cxx_compiler():
     return compiler
 
 
+def _get_compiler_launcher():
+    launcher = os.environ.get("PRIMUS_TURBO_COMPILER_LAUNCHER")
```
RuibinCheung: We don't want to add too many environment variable unless it is necessary. I proposed a plan that you can enable compiler launcher if ccache is installed.

RuibinCheung (changes requested):

tools/build_ext.py
jasainio: Done in `8ff0ea0` -- replaced the env var approach with auto-detection via `shutil.which()`. It now checks for `sccache` first, then `ccache` on PATH. No env var needed.

RuibinCheung (approved):
LGTM

PR #398, opened 2026-06-28
[feat] add mxfp4 triton grouped gemm support

kyle-256:
# Description

Adds an MXFP4 (E2M1) grouped GEMM op for MoE-style training on AMD CDNA4
(gfx950), with full autograd (forward + dgrad + wgrad) on the Triton backend.
The activation and weight quantization is fused so each operand's bf16 tensor is
read once and emits both the row-wise (fwd/dgrad) and col-wise (wgrad/dgrad)
MXFP4 operands, with the per-group M zero-padding and offsets computed on-GPU
(no D2H sync), making the forward CUDA-graph capturable. The MXFP4 training
recipe's Random Hadamard Transform on the backward operands is made bit-exact
deterministic.

Fixes # (N/A)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add `grouped_gemm_fp4` (MX_BLOCKWISE, E2M1, NT-only, Triton backend) with an
  autograd Function implementing fwd / dgrad / wgrad, paired RHT recipes per GEMM.
- Fuse the grouped **activation** dual-quant into one CUDA/HIP kernel: a single
  bf16 read produces the tight-M row-wise operand and the 128-padded per-group
  col-wise operand; pad offsets computed on-GPU (no bf16 scatter, no D2H sync).
- Batch the **weight** dual-quant over the 3D `(G, N, K)` weight in one launch
  (`blockIdx.z` per group) — removes the Python per-group loop and the
  full-weight transpose/contiguous copy.
- Make the row-wise-RHT quant bit-exact deterministic: fuse `ds_swizzle_b32` +
  `s_waitcnt lgkmcnt(0)` into one memory-clobbering asm so the reduction cannot
  latch stale lanes from in-flight LDS stores.
- Skip the redundant per-pass `warp_reduce_max_8` on the 2D-block scale path
  (the tile-wide `warp_reduce_max_64` already yields the same max) — bit-exact,
  cuts VALU on the VALU-bound weight quant.
- Use fp32 accumulation in the grouped-GEMM test reference (correct golden for a
  low-precision product comparison).
- Add an fp4 grouped-GEMM test suite aligned 1:1 with the fp8 suite (mx_blockwise
  sweep, zero-length groups, CUDA-graph, deterministic) plus an fp4-specific
  unaligned-N/K padded-contraction case.

# Checklist:

### DIFF
--- a/primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp4_impl.py
+++ b/primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp4_impl.py
@@ -1,0 +1,131 @@
+###############################################################################
+# Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
+#
+# See LICENSE for license information.
+###############################################################################
+
+"""MXFP4 grouped GEMM dispatch (Triton-only, gfx950).
+
+Thin ``torch.library`` custom-op wrappers around the Triton MXFP4 grouped
+kernels. Unlike the FP8 path (which fans out to CK / hipBLASLt / FlyDSL), MXFP4
+grouped GEMM currently has a single Triton block-scaled backend, so the ops call
+it directly. The forward op over-allocates the output to the (padded) input rows
+and the caller slices ``[:total_m]``; ``group_offs_out`` packs each group tight.
+"""
+
+import torch
+
+from primus_turbo.pytorch.core.low_precision import float4_e2m1fn_x2
+from primus_turbo.triton.grouped_gemm.grouped_gemm_fp4_kernel import (
+    grouped_gemm_mxfp4_triton_kernel,
+    grouped_gemm_mxfp4_variable_k_triton_kernel,
+)
+
+_torch_custom_op_wrapper = torch.library.custom_op
+
+
+@_torch_custom_op_wrapper("primus_turbo::grouped_gemm_fp4_impl", mutates_args=(), device_types="cuda")
+def grouped_gemm_fp4_impl(
+    a: torch.Tensor,
+    b: torch.Tensor,
+    a_scales: torch.Tensor,
+    b_scales: torch.Tensor,
+    group_offs: torch.Tensor,
+    N: int,
+    K: int,
+    num_cu: int | None,
+    out_dtype: torch.dtype,
+    group_offs_out: torch.Tensor | None = None,
+) -> torch.Tensor:
+    """Forward (NT): C[g] = A[g] @ B[g]^T.
+
+    a (total_M, K/2) fp4, a_scales (total_M, K/32) e8m0,
+    b (G, N, K/2) fp4, b_scales (G, N, K/32) e8m0. K is logical.
+    """
+    return grouped_gemm_mxfp4_triton_kernel(
+        a,
+        a_scales,
+        b,
+        b_scales,
+        group_offs,
+        N,
+        K,
+        group_offs_out=group_offs_out,
+        out_dtype=out_dtype,
+        num_cu=num_cu,
+    )
+
+
+@grouped_gemm_fp4_impl.register_fake
+def grouped_gemm_fp4_impl_meta(
+    a: torch.Tensor,
+    b: torch.Tensor,
+    a_scales: torch.Tensor,
+    b_scales: torch.Tensor,
+    group_offs: torch.Tensor,
+    N: int,
+    K: int,
+    num_cu: int | None,
+    out_dtype: torch.dtype,
+    group_offs_out: torch.Tensor | None = None,
+) -> torch.Tensor:
+    assert a.dim() == 2, f"a must be 2D, got {a.shape}"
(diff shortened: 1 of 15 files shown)


tests/pytorch/ref/gemm_ref.py
```diff
@@ -11,12 +11,15 @@
...
+    # NOTE: the per-group matmul is accumulated in fp32 (operands upcast, result
+    # cast back to ``a.dtype``): an fp32-accumulated golden is the right reference
+    # for a low-precision (fp4/fp8) product comparison.
     seg_lens = seg_lens.cpu().numpy()
     out = []
     start = 0
     for i, size in enumerate(seg_lens):
         rhs = b[i, :, :].t() if trans_b else b[i, :, :]
-        out.append(a[start : start + size, :] @ rhs)
+        out.append((a[start : start + size, :].float() @ rhs.float()).to(a.dtype))
```
RuibinCheung, line 22: Please don't change reference precision.

primus_turbo/pytorch/ops/grouped_gemm_fp4.py
```diff
@@ -0,0 +1,253 @@
...
+    group_offs_from_lens,
+)
+from primus_turbo.pytorch.kernels.quantization.quantization_impl import (
+    grouped_quantize_mxfp4_impl,
+)
+from primus_turbo.pytorch.ops.quantization import quantize_fp4_with_trans
+
+__all__ = ["grouped_gemm_fp4"]
+
+_PAD = 128  # per-group M padding for the variable-K wgrad (BLOCK_SIZE_K).
```
RuibinCheung: Use this variable( https://github.com/AMD-AGI/Primus-Turbo/blob/a04a233cbfb468dbe21600cbf9db70953428b25c/primus_turbo/pytorch/core/low_precision.py#L100) to replace this magic number.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp4_impl.py
```diff
@@ -0,0 +1,131 @@
...
+from primus_turbo.triton.grouped_gemm.grouped_gemm_fp4_kernel import (
+    grouped_gemm_mxfp4_triton_kernel,
+    grouped_gemm_mxfp4_variable_k_triton_kernel,
+)
+
+_torch_custom_op_wrapper = torch.library.custom_op
+
+
+@_torch_custom_op_wrapper("primus_turbo::grouped_gemm_fp4_impl", mutates_args=(), device_types="cuda")
+def grouped_gemm_fp4_impl(
```
RuibinCheung, line 250: Reference Grouped GEMM FP8, use dispatcher to control backend.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp4_impl.py
```diff
@@ -0,0 +1,131 @@
...
+    assert b.dtype == float4_e2m1fn_x2, f"b must be fp4, got {b.dtype}"
+    assert out_dtype in (torch.float16, torch.bfloat16)
+    # Output over-allocated to padded input rows; caller slices [:total_m].
+    return torch.empty((a.shape[0], N), device=a.device, dtype=out_dtype)
+
+
+@_torch_custom_op_wrapper(
+    "primus_turbo::grouped_gemm_fp4_variable_k_impl", mutates_args=(), device_types="cuda"
+)
+def grouped_gemm_fp4_variable_k_impl(
```
RuibinCheung: Keep API consistent with grouped gemm FP8 variable k.

primus_turbo/pytorch/kernels/grouped_gemm/grouped_gemm_fp4_impl.py
```diff
@@ -0,0 +1,131 @@
...
+        N,
+        K,
+        group_offs_out=group_offs_out,
+        out_dtype=out_dtype,
+        num_cu=num_cu,
+    )
+
+
+@grouped_gemm_fp4_impl.register_fake
+def grouped_gemm_fp4_impl_meta(
```
RuibinCheung, line 334: Keep API consistent with grouped gemm FP8.

RuibinCheung (changes requested):

tests/pytorch/ref/gemm_ref.py
kyle-256: The .float() here is deliberate — I'd prefer to keep it. The kernel under test accumulates in fp32 (MFMA), so the golden must accumulate in fp32 too: both sides consume the same quantized fp4 operands, and we want the tolerance to measure the kernel's error, not the reference's. A bf16-accumulated reference is strictly less accurate than the kernel it checks, which turns the comparison into a test of the reference's rounding. (Same reasoning the fp8 path relies on — this grouped_gemm_ref is shared by both.) The result is still cast back to a.dtype, so the returned golden dtype is unchanged.

RuibinCheung (approved):
LGTM

PR #405, opened 2026-07-03
fix(triton): scope AMD gfx950 compiler knobs to GEMM kernel launches

kyle-256:
# Description

The AMD Triton knobs (`use_async_copy` / `scalarize_packed_fops` /
`use_block_pingpong`) were set process-globally once (guarded by a
`_KNOBS_SET` flag) and never reverted. After any turbo GEMM / grouped-GEMM ran,
these knobs leaked into **unrelated** Triton kernels running in the same
process (e.g. the training loop): they raise shared-memory (LDS) usage — fine
for a standalone op, but enough to OOM during training — and regress some other
Triton kernels.

These knobs are **compile-time only**: they affect codegen and only need to be
active while a kernel is JIT-compiled on its first launch (Triton caches the
compiled binary, so later launches are unaffected). This PR keeps the knob
flipping inside the GEMM / grouped-GEMM entry points but wraps every entry
point with a new `scoped_amd_knobs` decorator that snapshots the knob state on
entry and restores it on exit.

Net effect: the default state is now **off**; turbo turns these knobs on only
while its own GEMM / grouped-GEMM kernels compile, so nothing else in the
process inherits them.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `primus_turbo/triton/utils/triton_knobs_helper.py`: drop the one-shot
  `_KNOBS_SET` global guard (knobs are now re-applied per call and restored by
  the surrounding scope); add `scoped_amd_triton_knobs()` context manager that
  snapshots/restores the AMD knobs (with `triton.knobs.amd` and env-var
  fallback), and a `scoped_amd_knobs` decorator wrapping it.
- Decorate all 16 GEMM / grouped-GEMM entry points with `@scoped_amd_knobs`:
  - `primus_turbo/triton/gemm/gemm_kernel.py` (1)
  - `primus_turbo/triton/gemm/gemm_fp8_kernel.py` (3)
  - `primus_turbo/triton/grouped_gemm/grouped_gemm_kernel.py` (2)
  - `primus_turbo/triton/grouped_gemm/grouped_gemm_fp4_kernel.py` (2)
  - `primus_turbo/triton/grouped_gemm/grouped_gemm_fp8_kernel.py` (8)
- The existing gfx942 layout-dependent `_set_amd_knobs(...)` calls are kept
  verbatim; they are now scoped (restored on function exit) too.

# Checklist:

### DIFF
--- a/primus_turbo/triton/gemm/gemm_fp8_kernel.py
+++ b/primus_turbo/triton/gemm/gemm_fp8_kernel.py
@@ -50,7 +50,10 @@ from primus_turbo.triton.utils.origami import (
     origama_hardware_info,
     origama_select_params,
 )
-from primus_turbo.triton.utils.triton_knobs_helper import set_triton_knobs_gfx950
+from primus_turbo.triton.utils.triton_knobs_helper import (
+    scoped_amd_knobs,
+    set_triton_knobs_gfx950,
+)
 
 # ═══════════════════════════════════════════════════════════════════════════════
 # AMD knobs helper (blockwise-specific)
@@ -255,6 +258,7 @@ def _fp8_persistent_gemm_kernel(
 # ═══════════════════════════════════════════════════════════════════════════════
 
 
+@scoped_amd_knobs
 def gemm_fp8_tensorwise_triton_kernel(
     a: torch.Tensor,
     a_scale_inv: torch.Tensor,
@@ -554,6 +558,7 @@ def _fp8_rowwise_persistent_gemm_kernel(
 # ═══════════════════════════════════════════════════════════════════════════════
 
 
+@scoped_amd_knobs
 def gemm_fp8_rowwise_triton_kernel(
     a: torch.Tensor,
     a_scale_inv: torch.Tensor,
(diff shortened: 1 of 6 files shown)


wenxie-amd (dismissed):

primus_turbo/triton/gemm/gemm_fp8_kernel.py
```diff
@@ -50,7 +50,10 @@
     origama_hardware_info,
     origama_select_params,
 )
-from primus_turbo.triton.utils.triton_knobs_helper import set_triton_knobs_gfx950
+from primus_turbo.triton.utils.triton_knobs_helper import (
+    scoped_amd_knobs,
+    set_triton_knobs_gfx950,
```
RuibinCheung: I think we don't need `set_triton_knobs_gfx950` anymore. Please move `set_triton_knobs_gfx950` into `scoped_amd_knobs`.

RuibinCheung (changes requested):

RuibinCheung (dismissed):
LGTM

RuibinCheung (approved):
LGTM

PR #408, opened 2026-07-07
fix(grouped_gemm): zero uncovered padding tail of grouped GEMM outputs

kyle-256:
fix(grouped_gemm): zero only the uncovered padding tail of grouped GEMM outputs

# Description

Grouped GEMM allocates its `forward`/`dgrad` output with `empty` and only writes
the covered rows `[0, sum(group_lens))`. When the input is over-allocated
(`M_total = a.size(0) > sum(group_lens)`, e.g. the fixed-capacity `moe_permute`
buffer), the trailing rows are left uninitialized and read back as stale memory
(`NaN`/`Inf`), non-deterministically corrupting MoE gradients.

Fix: keep the `empty` allocation (no full-tensor memset) and add a lightweight
Triton post-op `grouped_gemm_output_tail_kernel` that zeros only the uncovered
`[sum(group_lens), M_total)` tail after the GEMM. It reads the boundary on device
(no CPU sync) and is a no-op when there is no padding. Runs in the shared impl
layer after dispatch, so it covers all four backends (Triton, FlyDSL, CK,
hipBLASLt) at once. Variable-K wgrad already writes its full output, so it is
skipped.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add `grouped_gemm_output_tail_kernel` (Triton) that zeros only the uncovered
  output tail; wire it into `grouped_gemm_impl` and `grouped_gemm_fp8_impl`
  (non-MX path) after backend dispatch.
- Add regression test `test_grouped_gemm_fp8_padded_tail_zeroed` (poisons the
  allocator, checks the padded tail is zero) across all four backends × dtype ×
  `trans_b`.

# Checklist:

### DIFF
--- a/tests/pytorch/ops/test_grouped_gemm_fp8.py
+++ b/tests/pytorch/ops/test_grouped_gemm_fp8.py
@@ -1068,3 +1068,78 @@ def test_grouped_gemm_fp8_blockwise_zero_group_lens():
     b_grad_snr = compute_snr(b_ref.grad, b.grad)
     print(f"BGrad-SNR: {b_grad_snr:.2f} dB")
     assert b_grad_snr > snr_threshold, "b_grad_snr too low"
+
+
+def _poison_alloc_pool(shape, dtype, device, sentinel, n=24):
+    """Fill and free caching-allocator blocks of ``shape`` with ``sentinel`` so a
+    subsequent same-shape allocation reuses a dirty (non-zero) block instead of a
+    fresh, driver-zeroed page. Lets us detect output regions the kernel never wrote."""
+    blocks = [torch.full(shape, sentinel, dtype=dtype, device=device) for _ in range(n)]
+    for x in blocks:
+        x.add_(0.0)
+    del blocks
+
+
+@pytest.mark.parametrize("ori_dtype", ORI_DTYPE_VALUES)
+@pytest.mark.parametrize("trans_b", TRANS_B_VALUES)
+@pytest.mark.parametrize(
+    "backend", [BackendType.CK, BackendType.HIPBLASLT, BackendType.TRITON, BackendType.FLYDSL]
+)
+def test_grouped_gemm_fp8_padded_rows_zero_init(ori_dtype, trans_b, backend):
+    """issue #397: the FP8 grouped-GEMM forward only writes output rows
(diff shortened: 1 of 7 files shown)


csrc/pytorch/grouped_gemm/ck_grouped_gemm.cpp
```diff
@@ -107,7 +107,7 @@ at::Tensor ck_grouped_gemm(at::Tensor &a, at::Tensor &b, at::Tensor &group_lens,
     const int32_t m  = transA ? a.size(1) : a.size(0);
     const int32_t n  = transB ? b.size(1) : b.size(2);
     const int32_t k  = transA ? a.size(0) : a.size(1);
-    at::Tensor    c  = at::empty({m, n}, at::dtype(out_dtype).device(at::kCUDA));
+    at::Tensor    c  = at::zeros({m, n}, at::dtype(out_dtype).device(at::kCUDA));
```
xiaobochen-amd: FYI: The cost of at:zeros are greater than at::empty

kyle-256: use a triton tail function instead of using at::zeros tensors.

tests/pytorch/ops/test_grouped_gemm_fp8.py
```diff
@@ -1066,3 +1005,78 @@ def test_grouped_gemm_fp8_blockwise_zero_group_lens():
...
+        x.add_(0.0)
+    del blocks
+
+
+@pytest.mark.parametrize("ori_dtype", ORI_DTYPE_VALUES)
+@pytest.mark.parametrize("trans_b", TRANS_B_VALUES)
+@pytest.mark.parametrize(
+    "backend", [BackendType.CK, BackendType.HIPBLASLT, BackendType.TRITON, BackendType.FLYDSL]
+)
+def test_grouped_gemm_fp8_padded_tail_zeroed(ori_dtype, trans_b, backend):
```
RuibinCheung, line 1088: I noticed that you also change `grouped_gemm_fp4` and `grouped_gemm`. Please add same unit tests for `grouped_gemm_fp4` and `grouped_gemm`(bf16)

RuibinCheung (changes requested):

RuibinCheung (approved):
LGTM

PR #419, opened 2026-07-15
feat(grouped_gemm): decouple FlyDSL grouped fp8 autotune from M

kyle-256:
# Description

The tensorwise (TW) and MX-blockwise FlyDSL grouped fp8 GEMM kernels used to cache
and re-race their autotune config **per `total_m` / `M_pad`**, cold-compiling a full
candidate set on every new token count. But the compiled launch is **M-generic** (M
is a runtime kernel arg), so re-tuning per M is pure waste, and tuning on the
(possibly tiny / skewed) live first-call M can pick a poor config for the served
range.

This PR keys the config race on the **static shape only** and times it on **synthetic
canonical tensors**, so it runs once per shape and is reused for every M. It also
removes the previously observed wgrad regressions and caps worst-case per-M regret
with data-driven (not hardcoded) tuning points.

Fixes # (n/a)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

**Tensorwise NT / NN / wgrad** (`primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py`)
- Drop `total_m` from the autotune cache key; select via O(1) regime rules (coarse
  buckets folded into the key) + an NK race on balanced synthetic tensors, keyed on
  `(op, N, K, G, dtype, regime[, i64])` and reused across all M.
- Time on synthetic canonical tensors: rebuild only the M-dependent operands
  (`a`/`out` for NT/NN, `lhs`/`rhs` for wgrad) at the canonical size, reuse the
  M-independent side (`b` / `out` scratch).
- **wgrad `short` threshold recalibrated by a config×M sweep**: `short = M_total <= 4096`
  (was an inherited `pm <= 1536`), the root cause of the prior +4.8% / +3.5% wgrad
  regressions. wgrad now times a single live tokens/group point (no hardcoded M grid)
  and races the full 4-wave candidate pool.

**MX-blockwise NT** (`primus_turbo/flydsl/grouped_gemm/mxfp8_grouped_kernel.py`)
- Split caches — `_GNT_AT_CACHE` keeps `M_pad` (workspace/capture bake per M_pad),
  the config race keys on the M-independent `cfg_key`.
- The swizzle winner is not strictly M-invariant, so the race times synthetic
  balanced tensors (`_canon_nt_targs`, reusing the M-independent b-side) at two
  canonical steady points `{2048, 8192}` and scores by geomean (range-robust),
  capping worst-case bench regret at ~2.1% (vs 3.3% for a single point),
  deterministically. Adopt over base only if geomean is >=1.5% faster.
- Removed dead `_balanced_mx_targs`. Added env override `PT_MXGG_CANON_M` (race
  points); `PT_MXGG_AUTOTUNE=0` and the TW `_PM_CANON_OVERRIDE` A/B hook are kept.

**Behavior** — no numerical change: candidates are bijective tile→CU swizzles of the
same math (autotuned output is bit-identical to the base cfg, rel-RMSE 0), guarded by
a per-candidate rel-RMSE < 2e-2 check. Only *which* config is cached and *when* it is
raced changes.

## Test results

Validated on **gfx950 / MI355X** (chi2798). The kernel math is unchanged; the config
race now runs on the new canonical path (tests do not set `PT_MXGG_AUTOTUNE=0`).

| Suite | Result |
|---|---|
| `test_grouped_gemm_fp8.py --deterministic-only -k "mx_blockwise and FLYDSL"` (10× bit-exact) | **64 passed, 0 fail** |
| `test_grouped_gemm_fp8.py -k "mx_blockwise and FLYDSL"` (3 shapes × E4M3/E5M2/HYBRID × bf16/fp16; fwd+dgrad+wgrad) | **18 passed** |
| Falsification `PYTORCH_NO_CUDA_MEMORY_CACHING=1` slice | green (no allocator-masked OOB) |

### No regression / worst-case regret
- **MX**: autotuned output vs base cfg (`PT_MXGG_AUTOTUNE=0`) **rel-RMSE = 0**;
  worst-case bench regret (chosen cfg vs per-M best over M∈{2048,4096,8192})
  **≤ 2.1%**, deterministic (single-point 4096 was 3.3%).
- **TW**: interactive same-process A/B (per-M vs NK, isolates thermal / neighbor-GPU
  contention) — **72 items, worst B/A = 1.002** (was 1.048 before the threshold
  recalibration). Correctness (bal/unbal offs × non-unit scale × fwd/dgrad/wgrad)
  worst relerr **1.66e-3**.

### Lint
- pre-commit (`ruff check` + `ruff format`, line-length 110) all Passed.

# Checklist:

### DIFF
--- a/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
+++ b/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
@@ -1484,7 +1484,6 @@ _NP_8WAVE_CANDS = ((256, 8, 4, 0), (256, 8, 8, 0), (256, 8, 4, 4))
 # observed crossover (deepseek-up K=7168 wins with this at large M, nothing smaller does).
 _NP_LARGE_K = 4096
 _NP_PM_CANON = (1024, 8192)
-_PM_CANON_OVERRIDE = None
 
 
 def _np_regime(trans_b, N, K, G, M_total):
@@ -1550,7 +1549,7 @@ def _autotune_np_dispatch(trans_b, N, K, G, out_fp16, cbsz, blgp, args, regime):
 
     a_live, b_i8, out_live = args[0], args[1], args[2]
     mps = []
-    for pm in (_PM_CANON_OVERRIDE,) if _PM_CANON_OVERRIDE else _NP_PM_CANON:
+    for pm in _NP_PM_CANON:
         M_c = G * pm
         a_c = (torch.randn((M_c, a_live.shape[1]), device=a_live.device) * 0.1).to(a_live.dtype)
         out_c = torch.empty((M_c, N), device=out_live.device, dtype=out_live.dtype)
@@ -3058,7 +3057,7 @@ def _autotune_wgrad_dispatch(OUT_M, OUT_N, G, out_fp16, cbsz, blgp, args, short,
 
     lhs_live, rhs_live = args[0], args[1]
     M_total = lhs_live.shape[0]
-    pms = (_PM_CANON_OVERRIDE,) if _PM_CANON_OVERRIDE else (max(1, M_total // G),)
+    pms = (max(1, M_total // G),)
     mps = []
     for pm in pms:
         M_c = G * pm
--- a/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
+++ b/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
@@ -1479,35 +1479,30 @@ def _balanced_group_offs(m_total, G, device):
     return offs.view(torch.int32)
 
 
-def _balanced_targs(args, m_total, G):
-    """args with the group_offs slot (index 5) replaced by a balanced m_total/G split,
-    for distribution-independent autotune timing."""
-    bal = _balanced_group_offs(m_total, G, args[2].device)
-    return args[:5] + (bal,) + args[6:]
-
-
-# Fixed 8-wave candidates shared by NT (fwd) and NN (dgrad): (BLOCK_M, num_xcd, group_m,
-# group_n). Chosen from measured per-shape winners across m_total/G in 1024..8192, 7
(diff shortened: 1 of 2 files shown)


primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
```diff
@@ -1548,40 +1543,63 @@ def mk(bm, xcd, gm, gn):
             i64_traverse=i64_tr,
         )
 
-    pm = args[6] // G
-    bm128_tiles = G * ((pm + 127) // 128) * ((args[7] + 255) // 256)
-    if not trans_b and bm128_tiles <= _num_cus():
+    if not trans_b and regime == 1:
         # small-M dgrad: BLOCK_M=128 doubles M-tiles, beats every bm256 swizzle here
         # (boundary sweep +5..31%, never loses) -> single config, no autotune.
+        if os.environ.get("PT_TW_AT_DEBUG", "0") == "1":
```
RuibinCheung: Please remove this debug code.

primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
```diff
@@ -1609,18 +1627,18 @@ def _ok():
...
-            t = _robust_time(l4, targs)
-            # Re-time `best` right next to it to cancel thermal drift between candidates.
-            tb_now = _robust_time(best, targs)
-            if t < tb_now * 0.985:
-                best, bt = l4, t
+            s = _score(l4)
+            bs_now = _score(best)  # re-score best right next to it to cancel thermal drift
+            if s is not None and s < bs_now * 0.985:
+                best, bs, best_cfg = l4, s, ("4wave", xcd4, gm4, gn4)
+    if os.environ.get("PT_TW_AT_DEBUG", "0") == "1":
```
RuibinCheung: Same as above.

RuibinCheung (changes requested):

RuibinCheung (approved):
LGTM

PR #425, opened 2026-07-22
fix: grouped fp8 wgrad memory fault on empty (0-token) expert groups

kyle-256:
The variable-K wgrad builds a per-tile SRD num_records that can go negative for an empty group (mg==0): the 4-wave whole-loop folds the tile row offset into the SRD base and subtracts it from the record count (mg*OUT - tile_off). Fed to the unsigned SRD clamp, that negative wraps to ~4GB records, so the descriptor stops bounding the load and the kernel faults ("Memory access fault by GPU ... Unknown"). fwd/dgrad are unaffected; the fault only surfaces once the routing histogram has empty groups (e.g. the load-balancing-free MoE warmup router).

Floor num_records at 0 (arith.maxsi) at both wgrad num_records sites: _wave4_do_tile_tn (the 4-wave kernel, the one that actually faults, via create_buffer_resource) and _wgrad_rebase (shared by the masked and persistent kernels). An empty group then yields num_records=0 so every load HW-clamps to 0 -> the empty group dW is the zero matrix. mg>=1 is a no-op. Covers all three wgrad kernels.

# Description

The FlyDSL fp8 variable-K grouped GEMM used for the MoE weight-gradient ("wgrad") triggers a GPU memory-access fault (`Memory access fault by GPU ... Reason: Unknown`, SIGABRT) whenever the routing histogram (`group_lens`) contains **empty (zero-length) expert groups** — e.g. the load-balancing-free MoE warmup router that collapses onto a few experts. The forward and dgrad kernels are unaffected; the identical shapes run clean on the Triton backend and a balanced (no-empty) histogram runs clean on FlyDSL.

Root cause: for an empty group `mg = m_end - m_start == 0`, the per-tile SRD `num_records` goes **negative**. The 4-wave whole-loop folds the tile's row offset into the SRD base and subtracts it from the record count (`mg*OUT - tile_off`), so an empty group yields `-tile_off`. Fed to the unsigned SRD clamp (`minui`/`create_buffer_resource`), that negative wraps to a huge unsigned value and clamps to `0xFFFFFFFF` (~4 GB records) — the descriptor stops bounding the load and the kernel reads out of bounds → fault.

Fix: floor `num_records` at 0 (`arith.maxsi(..., 0)`) so an empty group yields `num_records == 0`; every load then HW-clamps to 0 and the empty group's `dW` is correctly the zero matrix. `mg >= 1` is a no-op (one scalar op per tile, outside the K-loop, no perf impact).

Fixes # (N/A — internal bug found via GPT-OSS-20B MoE training warmup)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `_wave4_do_tile_tn` (the 4-wave wgrad kernel — the one that actually faults, via `create_buffer_resource`): clamp `a_nrec`/`b_nrec` with `arith.maxsi(..., 0)`.
- `_wgrad_rebase` (shared by the masked and persistent wgrad kernels): same defensive `arith.maxsi(..., 0)` clamp on `a_nrec`/`b_nrec` so all three wgrad kernels uniformly guarantee a non-negative record count.

# Checklist:

- [ ] I have made corresponding changes to the documentation *(N/A — internal kernel bug fix, no user-facing docs)*
- [x] My changes generate no new warnings *(pre-commit ruff check/format pass)*

---

**Verification** (MI355X / gfx950, `grouped_gemm_fp8` TENSORWISE e4m3):

- Repro: GateUP shape K=2880 / N=5760 / G=32 with 13 empty experts — faults (exit 134) before the fix, all cases pass after.
- Per-kernel: a direct-call test bypasses the autotune dispatch and forces each of the 4-wave / persistent / masked wgrad kernels with 4 empty groups — all three run fault-free and match a torch reference (rel_err 0.0028); non-empty groups show no regression.

### DIFF
--- a/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
+++ b/primus_turbo/flydsl/grouped_gemm/gemm_fp8_grouped_kernel.py
@@ -1096,8 +1096,8 @@ def _wgrad_rebase(A, B, m_start, m_end, OUT_M, OUT_N, F8_IR_t):
     a_base = arith.index_cast(T.index, m_start) * arith.index(OUT_M)
     b_base = arith.index_cast(T.index, m_start) * arith.index(OUT_N)
     mg = arith.index_cast(T.index, m_end) - arith.index_cast(T.index, m_start)
-    a_nrec = mg * arith.index(OUT_M)
-    b_nrec = mg * arith.index(OUT_N)
+    a_nrec = arith.maxsi(mg * arith.index(OUT_M), arith.index(0))
+    b_nrec = arith.maxsi(mg * arith.index(OUT_N), arith.index(0))
     gA = make_fp8_buffer_tensor_rebased(A, F8_IR_t, a_base, a_nrec)
     gB = make_fp8_buffer_tensor_rebased(B, F8_IR_t, b_base, b_nrec)
     a_div = fx.logical_divide(gA, fx.make_layout(1, 1))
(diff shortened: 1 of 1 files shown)


RuibinCheung (approved):
LGTM

PR #430, opened 2026-07-27
feat(flydsl): support stochastic rounding in mxfp4 dual quant

kyle-256:
# Description

Route `use_sr` (stochastic-rounding) MXFP4 quant recipes through the FlyDSL dual
quant instead of the HIP C++ fallback. SR uses the hardware
`v_cvt_scalef32_sr_pk_fp4_f32` converter with a per-launch seed (mirroring the C++
`global_sr_counter`) and a grid-unique per-micro-block seed. Output is unbiased,
not bit-exact (SR is random by design); the non-SR path stays bit-exact vs HIP.

Fixes # (N/A)

# Type of change

# Changes

- `mxfp4_quant_kernel.py`: SR in the fused dual cast (2D dense + batched-3D).
  Adds `_next_sr_seed` / `_sr_hash`, a `seed` arg selecting the SR converter, and
  per-micro-block seeds; `dual_eligible` / `dual3_eligible` no longer exclude SR.
- `mxfp4_grouped_quant.py`: same for the grouped dual (`row_sr` / `col_sr`).
- `quantization_impl.py`: drop the `not use_sr` guards in the 3 FlyDSL routing
  gates and pass `row_sr` / `col_sr` from the recipes.

Two converter quirks handled: the SR op's `dst_sel` chaining corrupts inner bytes
(pack each pair into byte 0 then OR-shift, matching C++); and the `llvm.amdgcn` SR
intrinsic takes the seed one bit lower than the raw asm (shift seed left by 1).

# Checklist

- [x] My changes generate no new warnings (ruff check + format clean)
- [ ] I have added tests (verified out-of-tree vs the C++ SR path on gfx950 —
  non-SR byte-exact, SR unbiased/random; no test added to the in-repo suite yet)

### DIFF
--- a/primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
+++ b/primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
@@ -26,10 +26,13 @@ from flydsl.expr.typing import T
 from flydsl.expr.typing import Vector as Vec
 
 from primus_turbo.flydsl.quantization.mxfp4_quant_kernel import (
+    _SR_COL_SALT,
     _compute_scale_native,
     _cvt_microblock_to_fp4,
     _microblock_amax,
     _microblock_vf,
+    _next_sr_seed,
+    _sr_hash,
 )
 from primus_turbo.flydsl.utils.gemm_helper import make_row_band_resource, xcd_remap_pid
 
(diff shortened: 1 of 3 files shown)


RuibinCheung (approved):
LGTM

PR #453, opened 2026-08-11
perf(flydsl): dense fp8 NT/NN/TN optimisations

kyle-256:
# Description

Ports the dense FP8 GEMM optimisations for the NN (dgrad) and TN (wgrad) layouts onto `main`. Both layouts of this GEMM sit on the board power limit, so the contest is energy per FLOP rather than cycles; every change here takes work off the memory side or out of the dispatch, and none of them touch the NT (fwd) tile or its ISA.

`main` had not moved `gemm_fp8_kernel.py` since this line forked, so that file transfers whole. `gemm_helper.py` is a three-way merge — `main`'s own additions there survive, and three primitives this kernel calls that had drifted on `main` come back with it: `StoreCPerTensor`'s `store_aux`, `S2RLoaderTr.load`'s `vmcnt`, and the pool addressing the whole-loop tile needs.

The campaign harness, probe scripts and the grouped/mxfp8 lines from the source branch are deliberately left behind — this PR is two files.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

Non-breaking: existing entry points, signatures and defaults are unchanged; this adds a new TN tile path and new autotune candidates.

## Changes

**TN (wgrad)**

- A 4-wave whole-loop macro tile (`_compile_dense_tn_wave4`), one tile per workgroup at one wave per SIMD, with a split-K window and its reduce, gated by `_tn_wave4_supported` and raced per shape against the existing 8-wave kernel.
- Tile-level tuning on that carrier: cross-XCD partial handoff published at device scope, wait-state placement between the body's instruction groups, macro-tile/LDS-pool geometry (square and rectangular) with the mfma B step that spreads fragment refills, and band selection.

**NN (dgrad)**

- `StoreCPerTensorPairN`: a 16x16 mfma row of one n-fragment spans 16 lanes, so a scalar 2 B store leaves as a 32 B request. Packing the fragment pair with `v_cvt_pk_bf16_f32` and exchanging it with the neighbouring lane through a `quad_perm` DPP — no LDS, no barrier — lets each lane hold two adjacent columns, so the pair leaves as one 64 B request and the store count halves. Even `N`, bf16 output only.
- Store issued per accumulator group at that group's last mfma instead of joining one four-group burst after the last one; the exposed part of a tile's write drain is the tail of that burst.
- Removed one redundant LDS drain: the b0 loader's trailing `lgkmcnt(0)` is already covered by the following a0 load's own drain before `c00` consumes b0.
- Bands and XCD placement raced in the per-shape autotune, including the 2D band the NT path already used. An M-band reuses B's N-stripe, so once that stripe outgrows one XCD L2 slice the hardware round-robin beats clustering; narrow and wide grids want opposite bands, so both families are offered.

**Helper**

- Three-way merge of `gemm_helper.py`; no primitive is removed or re-signed, only extended.

## Measurements

MI355X (gfx950), one card, both arms materialised from git into isolated package
copies and md5-checked before the run, warmed once, then three interleaved passes;
medians below. Shapes are the forward `(M, K, N)` of `Y[M,N] = X[M,K] @ W[N,K]^T`:

(16384,  3072,  9216)   (16384,  3072,  3072)   (16384,  3072, 12288)
(16384, 12288,  3072)   (32768,  3072, 21504)   (32768, 15360,  3072)

Each metric is the geomean over the six cells of `torch._scaled_mm` time / ours.
NN and TN hand the reference a repacked operand and are not charged for the repack,
so those two are measured against a deliberately optimistic baseline; NT is the one
layout where neither side repacks (`W[N,K].t()` is already column-major).

| metric | `main` | this PR |
|---|---|---|
| `nt_gm` (NT / forward) | 0.958 | **1.085** |
| `nn_gm` (NN / dgrad)   | 0.937 | **1.068** |
| `tn_gm` (TN / wgrad)   | 0.815 | **1.057** |

Spread over the three passes was at most 0.71% on any arm. Correctness: SNR > 28 dB
on every cell; the accumulate (`beta=1`) epilogue is checked separately per store
class — output into a zeroed buffer is bit-identical to `beta=0`, and with `A = 0`
and a distinct value per element in `out` the result is bit-identical to that input.

The denominator is given row-major x column-major operands with no repack charged, so it is an optimistic baseline.

# Checklist:

- [ ] I have made corresponding changes to the documentation (n/a — no user-facing API or docs change)
- [x] My changes generate no new warnings (`ruff check` and `ruff format --check` clean; pre-commit passes)

### DIFF
--- a/primus_turbo/flydsl/gemm/gemm_fp8_kernel.py
+++ b/primus_turbo/flydsl/gemm/gemm_fp8_kernel.py
@@ -2583,7 +2583,7 @@ def _dense_nt_wave4_tile(
 
 
 @functools.lru_cache(maxsize=128)
-def _compile_dense_nt_wave4(
+def _compile_dense_wave4(
     M: int,
     N: int,
     K: int,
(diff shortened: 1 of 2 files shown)


RuibinCheung (approved):
LGTM

PR #456, opened 2026-08-13
feat(moe): fused MXFP8 mega MoE (dispatch+FC1, FC2+combine)

Xiaoming-AMD:
## What this adds

The bf16 mega MoE hides its expert all-to-all by fusing the comm into the same grid as the FC
GEMMs. This is the MXFP8 counterpart: pool operands are quantized to per-1x32-block E8M0 so the
GEMMs run on the scaled MFMA (`v_mfma_scale_f32_16x16x128_f8f6f4`), which both halves the bytes the
all-to-all moves and doubles the MFMA rate.

Measured on MI355X at the DSv3 shape (hidden 7168, inter 2048, 256 experts, topk 8, 8192 tokens),
against the existing bf16 mega MoE on the same node and in the same session, each arm the mean of
two reps with a cold `flyc` cache per rep:

| Routing | fp8 fwd+bwd | bf16 fwd+bwd | fwd+bwd | fwd-only | bwd-only |
|---|---|---|---|---|---|
| load_balanced | 13.77 ms | 20.38 ms | 1.48x | 1.42x | 1.57x |
| round_robin | 13.23 ms | 19.58 ms | 1.48x | 1.42x | 1.57x |

The backward is where it pays off most (1.57x against 1.42x on the forward), and the ratio is
unchanged under heavily skewed routing, so the win does not depend on balanced experts. Repeat
spread on this bench reaches ~1.5% within one session and more across sessions, so anything smaller
than that is noise.

## Accuracy

An end-to-end training step against a bf16 reference. Both arms use
`deepseek_v3-BF16-pretrain.yaml`, so the whole model is bf16 and the only difference is whether the
MoE experts run through MegaMoE in mxfp8. DeepSeek-V3 at 4 layers, 8x MI355X, EP 8, global batch
512, 50 iterations, same seed.

![bf16 reference vs bf16 + MegaMoE mxfp8 experts: loss curves and the per-iteration difference]
image

| iteration | bf16 reference | bf16 + MegaMoE mxfp8 | difference |
|---|---|---|---|
| 1 | 12.0143 | 12.0148 | -0.0004 |
| 10 | 10.1209 | 10.1238 | -0.0029 |
| 20 | 7.8481 | 7.8502 | -0.0022 |
| 30 | 6.0425 | 6.0221 | +0.0204 |
| 40 | 4.9202 | 4.8988 | +0.0214 |
| 50 | 4.5221 | 4.5072 | +0.0149 |

Final difference **+0.0149, or 0.33% of the reference loss**; mean absolute difference 0.0121, max
0.0290, and the sign of the difference changes 5 times over the 50 iterations. That last number is
the one that matters: a systematic accuracy loss would hold one sign and grow, so what is left reads
as quantization jitter rather than drift.


zhenhuang12:
Hi @Xiaoming-AMD 
Do we have plans to unify the common abstractions, such as `prims.py, barrier.py, dispatch_prologue, sym_buffer`

Xiaoming-AMD:
> Hi @Xiaoming-AMD Do we have plans to unify the common abstractions, such as `prims.py, barrier.py, dispatch_prologue, sym_buffer`

Planned, but not in this PR. I'd rather do it as a follow-up that looks at bf16 and mxfp8 together.

zhenhuang12 (approved):

wenxie-amd (approved):

PR #458, opened 2026-08-13
feat(attn): support sbhd & thd attn for meta and gpt-oss

kyle-256:
# Description

Adds a **FlyDSL attention backend** for gfx950 (MI355X) and the multi-backend plumbing to
reach it: a forward, a deterministic backward, sparse-MLA, and the dispatcher that picks
between FlyDSL and aiter per shape.

The backward is the substantial part. It is a single KV-outer pass that also emits dQ (5
GEMMs instead of the split pair's 7), with the dQ contribution of each kv band going to a
bf16 split-K workspace that a reduce folds in fixed band-ascending fp32 order — no atomics,
bitwise reproducible. It covers every causal shape the kernels admit: square and
rectangular, sliding window, THD document packing, and long context (where the workspace is
walked in band groups so its footprint follows a cap rather than the square of the context).

Once the fused path was faster on every shape the split pair served, the split pair was
deleted — `build_flash_attn_bwd_dq_module` and everything that only fed it, ~2000 lines.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

**FlyDSL attention kernels** (new files under `primus_turbo/flydsl/attention/`)

- Forward: SBHD-native and THD varlen, causal and sliding-window, D in {64, 128}, GQA, and a
  learned per-head attention sink folded into the softmax denominator.
- Backward: deterministic 16x16x32, dQ fused into the KV-outer pass. Bottom-right causal,
  rectangular shapes, sliding windows, ragged/block-causal packing, and a band-group walk
  that bounds the dQ workspace at long context.
- Sparse-MLA (DeepSeek-V4) forward and backward.

**Multi-backend dispatch**

- `attention_impl.py`: `KernelBackend` entries for FlyDSL and aiter behind an
  `AutoKernelDispatcher`, mirroring the GEMM convention. Forward and backward must agree on a
  backend, so the op layer resolves once per call and hands the enum to the autograd
  Function, which carries it on ctx.
- `flash_attn_func` / `flash_attn_varlen_func` are one dispatching Function each rather than
  one Function per backend.
- `PRIMUS_TURBO_ATTN_BACKEND` / `PRIMUS_TURBO_SPARSE_ATTN_BACKEND` and
  `GlobalBackendManager.set_attn_backend`, same per-precision format as the GEMM knobs.
- FlyDSL eligibility asks the tensors what it needs — whether the `[s, b, h, d]` view is
  contiguous — instead of comparing a storage-order name, so a single batch is no longer
  pushed off it over a distinction that does not exist at `b == 1`.
- The four FlyDSL entry points are registered as torch custom ops with fake kernels, so a
  `torch.compile(fullgraph=True)` caller does not trace into their JIT build.

**Deletions**

- The split (Q-outer) backward pair, now that nothing reaches it: ~2000 lines, including 16
  builder knobs no caller ever set and the `Q_PAIR` machinery behind them.

**Benchmarks**

- The FlyDSL benchmarks are folded into `bench_attention_turbo.py` as options
  (`--backend`, `--layout`, `--varlen`) rather than separate scripts; shape tables live in
  `config.py` with the rest.

## Fixes found along the way

- The aiter backward allocated `dk`/`dv` with the **query** sequence length, so a rectangular
  shape asked aiter for the wrong `dk` and it refused. Square shapes hide it, which is every
  shape the suite had.
- The backend-fallback warning failed `torch.compile`: dynamo cannot trace a `logging.Logger`
  call, so any shape whose default backend was ineligible failed to compile — 80 of the 200
  cases in `tests/pytorch/modules/test_attention.py`.
- Sparse-MLA passed the dispatcher a bare `BackendType` where it takes a `BackendChoice`, so
  every call raised `AttributeError`. That path had never run.
- Two references were top-left causal (`is_causal`), which is the kernels' bottom-right
  masking only while `seqlen_q == seqlen_k`.
- A window narrower than a q block returned NaN dQ (the dQ reduce took a whole q block's band
  range while the body had written half of it); such windows now go to aiter.

# Checklist:


benchmark/ops/training/bench_flydsl_flash_attn_bwd.py
```diff
@@ -0,0 +1,185 @@
+###############################################################################
```
RuibinCheung: Megre these file into existence benchmark files

primus_turbo/pytorch/ops/attention/flash_attn_interface.py
```diff
@@ -220,6 +237,71 @@ def backward(ctx, dout, *args):
         )
 
 
+class FlydslFlashAttnFunc(torch.autograd.Function):
```
RuibinCheung: Remove `FlydslFlashAttnFunc` and add flash attn dispatcher to support dispatch to aiter and flydsl backend.

primus_turbo/pytorch/ops/attention/flash_attn_interface.py
```diff
@@ -582,6 +709,90 @@ def backward(ctx, dout, *args):
         )
 
 
+class FlydslFlashAttnVarlenFunc(torch.autograd.Function):
```
RuibinCheung: Remove `FlydslFlashAttnVarlenFunc` and add flash attn varlen dispatcher to support dispatch to aiter and flydsl backend.

wenxie-amd (approved):

PR #461, opened 2026-08-17
chore(flydsl) drop env knobs

kyle-256:
# Description

Removes the leftover experiment env knobs from the FlyDSL GEMM kernels, and fixes a silent
wrong-answer bug found while validating that removal.

Seven `PT_*` / `MX_*` variables were still being read at kernel-build time across
`gemm_mxfp8_kernel.py` and `grouped_gemm_fp8_kernel.py` — nine reads in all. Every one of
them defaulted to the value production actually runs, so each read is a constant and the
other side of every branch it guarded was unreachable. They are hardcoded to those values
and the dead branches deleted.

The bug is unrelated to the knobs but sat on the path the validation run exercised. Grouped
FP8 tensorwise `Format.HYBRID` returned NaN and Inf from wgrad — not a low-SNR result, a
wholly destroyed gradient. It reproduces on `main`.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

**Env knobs hardcoded to their shipped values** (`2119b4a8`)

| Variable | Default that shipped | Result |
| --- | --- | --- |
| `MX_DISABLE_NT_GN` | unset = don't disable | branch and its docstring sentence dropped |
| `PT_NN_WSWZ` | `1` | `wswz=True` at both call sites |
| `PT_NO_G4W` | `0` | condition collapses to `if not trans_b:` |
| `PT_QUIET_G4W` | `0` | the warning is unconditional; `import warnings` moved to module scope |
| `PT_WL_VMCNT_MODE` | `partial` | variable gone, two conditions collapse, and the constant leaves the asm cache key |
| `PT_WL_2BPOOL` (×3) | `1` | 2-pool pinned on, which fixes eight downstream branches |
| `PT_CS` | `1024` | written directly |

Pinning `PT_WL_2BPOOL` takes the 1-pool fallback with it: `b0_extra_buf` is non-`None` at
both call sites, so its `None` handling comes out of `_wholeloop_tile_3buf` — the prologue's
third prime, the pool-2 buffer tuple, and the `soff0_b` slot count all stop being
conditional. `import os` is no longer needed in either file.

**Mixed-fp8 wgrad correctness** (`74a95257`)

`_get_gg_fp8_tw_vk_config` and `_get_gg_fp8_rw_vk_config` now pin `BLOCK_K=128` when the two
operands have different fp8 dtypes, and stop accepting a 64-wide origami candidate for that
case. This is the constraint the forward config has always carried.

## Fixes found along the way

**Grouped FP8 `Format.HYBRID` wgrad returned NaN/Inf.**

Below `BLOCK_K=128`, Triton selects an MFMA whose opcode bakes a single fp8 format for both
operands. With one operand e4m3 and the other e5m2 the e4m3 side is decoded as e5m2, and its
`0x7c`–`0x7f` codes are exactly e5m2's Inf/NaN encodings — which is why the failure produces
both, rather than merely wrong magnitudes.

Only wgrad was affected, because only wgrad reached a narrow K:

| Path | `BLOCK_K` |
| --- | --- |
| forward / dgrad (`_get_gg_fp8_tw_fwd_config`) | pinned 128; origami accepted only at `ok == blk_k` |
| **wgrad (`_get_gg_fp8_{tw,rw}_vk_config`)** | defaulted to **64**, and `ok in (64, 128)` let origami keep 64 |

Measured on gfx950 with `B=8, M=1024, N=2048, K=1536`:

- `BLOCK_K=64` is wrong at 256×256, 128×128 and 64×64 — tile shape does not matter.
- `BLOCK_K=128` is correct at 128×128, 256×128, 128×256 and 256×256.
- Before: `b.grad` held 2 928 211 NaN and 4 815 277 Inf. After: none, and its finite absmax
  goes from 183 back to 204 against a reference absmax of 207.
- E4M3-only and E5M2-only were correct throughout, which is what kept this hidden: the
  operands are uniform in both, and `Format.HYBRID` is the only case that mixes them.

A standalone Triton reproducer (a bare `tl.dot` on two fp8 tensors of different dtype) shows
the same at `BLOCK_K=32`, so the miscompile is upstream of this repo; pinning `BLOCK_K`
keeps us off it.

The non-gfx950 branch of both config functions is deliberately unchanged. It also defaults
to 64, but 256×256×128 double-buffered does not fit gfx942's 64 KB LDS, and there is no part
here to verify a replacement on.

## Validation

- `tests/pytorch/ops/test_grouped_gemm_fp8.py` + `tests/pytorch/ops/test_gemm.py` were
  981 passed / 1 failed before this branch. The single failure is
  `test_grouped_gemm_fp8_tensorwise[...Format.HYBRID...]`, i.e. exactly the bug fixed here,
  and it reproduces identically on `main`.
- `ruff check` and `ruff format --check` are clean over `primus_turbo/`.
- Every one of the 46 top-level names the branch touches still has a reference; no symbol was
  left orphaned by the branch deletions.

# Checklist:


RuibinCheung (dismissed):
LGTM

RuibinCheung (approved):
LGTM

RuibinCheung (changes requested):
PTAL

RuibinCheung (approved):
LGTM

PR #462, opened 2026-08-18
feat(attn): add HipKittens gfx950 attention backend for head dim 64/128 causal + SWA

wenxie-amd:
# Description
This PR brings the HipKittens gfx950 attention kernels into Primus-Turbo as a dense attention
backend, reachable through the ordinary `flash_attn_func` path and the existing dispatcher.
HipKittens is a header-only CDNA4 tile library; its attention kernels are written against
gfx950 ISA features (`__builtin_amdgcn_mfma_f32_16x16x32_bf16` and friends) that gfx942 does
not have. Rather than ship a separate gfx950-only extension, each kernel body is guarded on
`__gfx950__` and asserts on any other device pass, so the four translation units compile into
every arch configuration and land in the same `_C` extension as everything else. What keeps
them off the wrong card is the `is_gfx950()` check in the capability gate, not the build.

Fixes # (issue)
## Type of change
- [x] Infra/Build change
- [ ] Code refactoring
## Changes
- Add HipKittens as a submodule at `3rdparty/hipkittens`, pinned at `a288366e`.
- Bring in the four attention kernels (`hk_attn_{fwd,bwd}_d{64,128}.cu`), each kernel body
  guarded on `__gfx950__` so they build for every arch and refuse to run on the wrong one.
- Add a torch adapter layer (`csrc/pytorch/attention/hk_attn.cpp`, `tensor_gl.h`) that converts
  tensors to `kittens::gl<>` descriptors, replacing HipKittens' own pybind11 entry points. Eight
  ops register into the shared `primus_turbo_cpp_extension` namespace -- no separate `.so`.
- Add the Python layer `primus_turbo/hipkittens/attention/gfx950/`, which owns the sequence
  padding, the dK/dV split sizing, and `hipkittens_attn_supported()` -- a capability gate that
  returns the reason it refuses, so a pinned backend can say what it rejected instead of
  silently handing the work to another backend.
- Register `BackendType.HIPKITTENS` as a dense forward/backward backend behind
  `attention_impl.py`, and wire both directions through `FlashAttnFunc`.
- Always take the split dq + dkdv backward. The fused single-pass variant stays in the kernels
  and is not dispatched to: over every shape whose workspace it fit, it ran 2.2x to 3.6x slower
  on the backward, and past that point its workspace passes the 4 GiB buffer-descriptor limit
  and the stores are silently dropped, so there is no larger shape where it wins.
- Extend `test_attention.py`: HIPKITTENS joins the backend axis of the shared 16-bit tests, so
  the cells it cannot take become assertions that it *refuses* them, plus two shape families of
  its own (ten rectangular/windowed meta configs and eight deduplicated pretrain head structures).
- Add `--backend hipkittens` to `bench_attention_turbo.py`, and add GPT-OSS-20B to the benchmark
  model table (the first head-dim-64 entry with a GQA group of 8 at 4096).
## Supported envelope
The kernels take **causal attention at head dim 64 and 128 only**, and a **left window (SWA)**
is supported at both. Everything outside this is refused by `hipkittens_attn_supported()` with a
reason rather than silently mishandled:
| | supported |
|---|---|
| Arch | gfx950 only (MI350X / MI355X) |
| dtype | bf16 only |
| Head dim | 64, 128 (`Dqk == Dv`, so no MLA) |
| Masking | causal only; non-causal refused |
| Window | left window (SWA) supported; right window must be 0 or -1 |
| Shape | `Sq <= Skv`, `Hq % Hkv == 0` (MHA / GQA / MQA) |
| Layout | dense 4-D SBHD, contiguous; no varlen / THD packing |
| Not implemented | dropout, bias, alibi, learned sink |
## Performance
MI355X (gfx950), bf16, sbhd, causal, `Sq == Skv`, no window. TFLOPS, effective (causal FLOPs
counted as half). The 18 shapes are the model shapes all three dense backends accept -- FlyDSL
sets the intersection; HipKittens itself takes 36 of the table's 90 cases. Bold is the best of
the three in that row.
| D | S | Hq/Hkv | B | aiter fwd | flydsl fwd | hk fwd | aiter bwd | flydsl bwd | hk bwd |
|---|---|---|---|---|---|---|---|---|---|
| 64 | 4096 | 64/8 | 1 | 622 | **989** | 985 | 303 | **757** | 619 |
| 64 | 4096 | 64/8 | 2 | 646 | **1040** | 1010 | 303 | **766** | 672 |
| 64 | 4096 | 64/8 | 4 | 653 | **1056** | 1019 | 303 | **772** | 701 |
| 64 | 8192 | 32/4 | 1 | 665 | 1044 | **1072** | 328 | **856** | 661 |
| 64 | 8192 | 32/4 | 2 | 691 | 1074 | **1097** | 327 | **867** | 723 |
| 64 | 8192 | 32/4 | 4 | 699 | 1101 | **1110** | 328 | **851** | 737 |
| 64 | 8192 | 64/4 | 1 | 704 | 1099 | **1109** | 328 | **841** | 693 |
| 64 | 8192 | 64/4 | 2 | 705 | 1104 | **1116** | 329 | **866** | 741 |
| 64 | 8192 | 64/4 | 4 | 704 | **1119** | **1119** | 327 | **853** | 738 |
| 128 | 4096 | 64/8 | 1 | 517 | 1011 | **1059** | **790** | 512 | 628 |
| 128 | 4096 | 64/8 | 2 | 520 | 1053 | **1063** | **795** | 700 | 656 |
| 128 | 4096 | 64/8 | 4 | 521 | 1061 | **1072** | **799** | 688 | 647 |
| 128 | 8192 | 64/8 | 1 | 552 | 1134 | **1168** | **921** | 810 | 705 |
| 128 | 8192 | 64/8 | 2 | 562 | 1139 | **1173** | **920** | 792 | 700 |
| 128 | 8192 | 64/8 | 4 | 566 | 1140 | **1176** | **929** | 755 | 693 |
| 128 | 8192 | 128/8 | 1 | 562 | 1139 | **1178** | **915** | 758 | 708 |
| 128 | 8192 | 128/8 | 2 | 565 | 1141 | **1179** | **956** | 753 | 692 |
| 128 | 8192 | 128/8 | 4 | 566 | 1140 | **1179** | **956** | 719 | 695 |
| | | **mean** | | 612 | 1088 | **1105** | 603 | **773** | 689 |

Forward is the win: fastest of the three on 15 of 18 shapes, and the three it loses are the
shortest kv in the set (D=64, S=4096), by 0.4% to 3.5%. Accuracy is 18/18 PASS at the harness
threshold (SNR > 40 dB against a torch reference).
Backward does not lead anywhere yet, and this is the known gap: FlyDSL is 6%-18% ahead at
D=64, and aiter is 25%-27% ahead at D=128. That is not the end state -- the backward will keep
being optimized in follow-up PRs, with the D=128 case first in line as the largest single item
left. What lands here is correct on every shape it accepts and already carries the forward win,
so it is worth having in tree while that work continues.

## Checklist:
- [x] I have made corresponding changes to the documentation
`tests/pytorch/ops/test_attention.py`: 10094 passed, 0 failed; the HipKittens subset
(`-k hipkittens`) is 3365 passed.


RuibinCheung (approved):
LGTM

PR #470, opened 2026-08-21
FlyDSL: pad-aware grouped-fp8 down-proj and native fused grad-accum for gpt-oss-20b MoE

kyle-256:
# Description

Native pad-aware FlyDSL grouped-fp8 path for the gpt-oss-20b MoE down-projection,
plus native fused gradient accumulation (Megatron `main_grad += wgrad`, beta=1).

The gpt-oss-20b down-proj GEMMs are non-128-aligned (deploy shape N=K=2880). On
`main`, such shapes fall back to a slow generic path. This PR adds a copy-free
pad-aware path (K→ceil128(K), weight N→ceil128(N)) that writes a tight output with
no host-side slice-back, and folds the wgrad fused grad-accum straight into
`main_grad` — for all three main_grad dtypes Megatron may allocate (fp32 / bf16 /
fp16) — removing the separate host `main_grad.add_()`.

Fixes # (N/A)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Native pad-aware grouped-fp8 down-proj (FlyDSL)**: fwd (NT) / dgrad (NN) / wgrad
  (variable-K TN) take real `n_real` / `m_real` extents so non-128-aligned N/K produce
  a tight, copy-free output (no slice-back on the host). Pad is unconditional at the
  quant/interface layer; the tight hint is only passed when the shape actually shrinks,
  so 128-aligned shapes are bit-exact no-ops (zero regression).
- **Native fused grad-accum (beta=1)** in the tight wgrad epilogue: accumulates
  directly into `main_grad`, eliminating the host `add_`. Store width mirrors
  `main_grad.dtype` and is dtype-aware for **fp32 / bf16 / fp16** (three-state
  `_out_ty`; cache key carries `out_fp16`/`out_fp32` to avoid kernel reuse collisions).
- **fp32 fast-CShuffle epilogue**: fp32 accumulate goes through the 128b coalesced
  CShuffle store (LDS-fitted at `epi_pad=0`); scalar store only as a fallback for N not
  a multiple of the CShuffle column tile.
- **Config equality for beta=1**: candidates are always raced at beta=0 into scratch and
  a `make(beta_is_one)` factory is returned, so the tuned config is beta-independent —
  no beta-specific config penalty.
- **Boundary-tile fault fix**: the both-last corner tile under tight + beta=1 read-back
  routed the 2-pool `(1,1)` geometry into a FlyDSL whole-loop codegen fault; it now takes
  the 3-pool `(2,1)` geometry via a compile-time `const_expr(c_tight)` (beta-independent,
  keeps beta=1 == beta=0 geometry).

## Benchmarks

TFLOP/s (higher is better), min-of-30, gfx950 / MI355X. **All FlyDSL: mine-flydsl vs
main-flydsl** — isolates the pad path net gain (same backend both sides).

All FlyDSL: mine-flydsl vs main-flydsl (pad net gain, with quant)
| case | fwd main→mine | bwd main→mine |
|---|---|---|
| gpt_oss-20b-Down\|B32\|M4096\|balanced | 1316→1421 (+8.0%) | 1830→1974 (+7.9%) |
| gpt_oss-20b-Down\|B32\|M4096\|unbalanced | 1312→1393 (+6.2%) | 1716→1897 (+10.5%) |

Non-128 shapes (gpt-oss-20b down-proj, Qwen/DeepSeek down-proj) gain across the board;
128-aligned GateUP shapes stay within ±DVFS noise (no regression). Fused beta=1 wgrad is
net-positive vs the unfused (beta=0 + host `add_`) alternative for all three main_grad
dtypes (measured +21% fp32 / +9% bf16 / +7% fp16 kernel read-back, but 13–47% faster than
unfused end-to-end).

# Checklist:

- [x] New and existing unit tests pass locally with my change

### DIFF
--- a/primus_turbo/pytorch/ops/grouped_gemm_fp8.py
+++ b/primus_turbo/pytorch/ops/grouped_gemm_fp8.py
@@ -88,12 +88,6 @@ def _grouped_gemm_fp8_variable_k_impl_wrapper(
     if not inplace_add_to_out:
         return grouped_gemm_fp8_variable_k_impl(*inputs, m_real=m_real, n_real=n_real, **options)
 
-    # Fused bgrad-accum: the wgrad accumulates straight into ``out`` (beta=1). Tight (N/K-pad)
-    # output accumulates natively in the FlyDSL fast CShuffle 128b epilogue (fp32-/pad-aware,
-    # scalar store only as fallback for a non-Cc-multiple N) -- no host add_ / branch. The
-    # accumulate target (Megatron's main_grad) may be fp32 while the weight -- and thus the
-    # dummy wgrad handed back to autograd -- is 16-bit; each backend reconciles the fp32 ``out``
-    # against the (16-bit) ``out_dtype`` internally, so ``out_dtype`` stays the weight dtype here.
     assert out is not None, "out is required when inplace_add_to_out is True"
     grouped_gemm_fp8_variable_k_accum_impl(*inputs, out=out, m_real=m_real, n_real=n_real, **options)
     return _get_dummy_wgrad(out.shape, out_dtype)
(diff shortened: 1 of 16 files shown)


primus_turbo/pytorch/ops/grouped_gemm_fp8.py
```diff
@@ -66,15 +67,12 @@ def _grouped_gemm_fp8_variable_k_impl_wrapper(
     default_backend: int,
     inplace_add_to_out: bool = False,
     out: Optional[torch.Tensor] = None,
+    m_real: Optional[int] = None,
```
RuibinCheung, line 70: Can we get m_real and n_real from out ?

RuibinCheung (approved):
LGTM

PR #468, opened 2026-08-21
perf(flydsl): tune the MXFP4 dual quant for gpt-oss-20b MoE on gfx950

zhitwang17:
## Summary

This PR optimizes the MXFP4 quantization path on top of Primus-Turbo PR #460 for GPT-OSS-20B MoE training on AMD Instinct MI355X/gfx950.

Main changes:

- Increase grouped dual-quant `BM` from 64 to 128 to amortize workgroup-level metadata scanning, indexing, SRD construction, and kernel-prologue overhead.
- Replace full-output initialization with selective tail-zeroing for batched dual quantization.
- Introduce the compact K256 wgrad contract, reducing per-expert colwise padding alignment from 512 to 256 across the quant producer, wgrad consumer, and runtime whole loop.
- Extend the runtime whole loop to handle zero-length groups, even pairs of 256-row blocks, and odd 256-row tails.
- Integrate the GEAK stacked grouped-Quant candidate, including N-grid collapse and direct colwise global-memory stores.
- Preserve PR #460's fused gradient-accumulation path (`beta=1` / `out`) while applying the K256 consumer contract.
- Add regression guards for padded-zero semantics, grouped colwise metadata, zero/even/odd K256 spans, and fused gradient accumulation.

## Performance

The final source was measured directly against `main@05f0f4f7` after PR #460 was merged. Testing used MI355X/gfx950, 24 representative GPT-OSS-20B EP1/G32 routing traces, and both production grouped-GEMM shapes. The full-op boundary includes Quant, fwd, dgrad, and wgrad.

Results are pooled from three independent paired sessions using the equal-session mean of route-weighted latencies, followed by the baseline/candidate ratio.

| Phase | PR #460 baseline | This PR | Speedup |
|---|---:|---:|---:|
| Quant | 3.2448 ms | 2.4467 ms | **1.326166x (+32.62%)** |
| Fwd | 1.6050 ms | 1.6115 ms | 0.995979x (-0.40%) |
| Dgrad | 1.5057 ms | 1.5114 ms | 0.996244x (-0.38%) |
| Wgrad | 1.6353 ms | 1.6045 ms | **1.019210x (+1.92%)** |
| Full public op | 8.1669 ms | 7.3506 ms | **1.111064x (+11.11%)** |

The three session-level ranges were:

- Quant: **1.3157x to 1.3344x**
- Full-op: **1.1076x to 1.1130x**

[...]

PR #460, opened 2026-08-17
perf(flydsl): tune the MXFP4 GEMM for gpt-oss-20b MoE on gfx950

kyle-256:
# Description

The FlyDSL MXFP4 dense and grouped GEMM kernels are retuned for the gpt-oss-20b MoE shapes on gfx950. The work was driven by a benchmark board that scores the MoE projections under **unbalanced expert routing**, which is where the previous kernels lost the most: a hot expert forced every group onto its tiling, so the wgrad collapsed as soon as the token distribution stopped being uniform.

Two changes carry almost all of the gain:

- **wgrad tile blocking is chosen from the per-group contraction, not one global config.** A short contraction writes G× more C for the same FLOP, so the store saturates and evicts the operand slabs it shares L2 with — a non-temporal store buys those hits back. Past one CU's worth of tiles the machine sits inside a single group, so its own operand rows must stay resident and the wide band wins instead.
- **A thinner NT candidate for the tall-narrow shapes** the MoE projections produce.

This branch is rebased onto `main` as a single commit. `main`'s own work on these files is preserved rather than overwritten — `StoreCPlain` now carries both the `beta_is_one` accumulate epilogue (grad accumulation fuse, #443) and the interleaved-column/span-drop store this work needs, and the `_out_ty` hoist from #440 is re-applied at both store sites of the restructured file. Only the MXFP4 kernels are taken from the tuning branch; the MXFP8/FP8 kernels it also touched are left to their own branches.

Fixes # (N/A)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [x] Code refactoring *(performance tuning; no API change)*

## Changes

- **`flydsl/grouped_gemm/grouped_gemm_mxfp4_kernel.py`** — wgrad tile blocking selected from the mean per-group contraction, with a second split on the per-group tile count; non-temporal C store for the thin-group arm; lane-resident group table (a tile's contraction bounds cost one `v_readlane` instead of a global load an occ=1 prologue cannot hide) replacing the serial G-wide compare tree; LPT walk order so the longest group is not exposed as a tail; band-cyclic XCD remap.
[...]

PR #474, opened 2026-08-21
flydsl: optimize mxfp4 quant kernel part2

kyle-256:
# Description

Rewrites the FlyDSL grouped MXFP4 dual quant for the gpt-oss-20b MoE regime on gfx950,
and fixes three correctness defects in the MXFP4 wgrad whole-loop that the rewrite and
the 256-row colwise contract exposed.

The quant was the one piece of the MoE step nobody had tuned: at the deployment regime
(EP1, G=32 local experts, per-expert M=4096, hidden 2880 -> 2944) it cost 0.90 ms against
a 4.75 ms GEMM, i.e. 16% of the step. It is now 0.45 ms (8.7%), with the GEMM untouched.

This PR overlaps #468, which tunes the same kernel. It adopts #468's 256-row per-group
colwise contract (producer, consumer and the runtime loop bound) on top of this quant,
so the two are not additive -- this supersedes #468's version of the kernel. Rebasing
onto current `main` will conflict in `mxfp4_grouped_quant.py`; the resolution is "keep
this branch's quant, keep the 256 contract".

Fixes # (no issue -- the wgrad defects were found while validating this branch; the
bit-exactness test added for #427 is what catches them)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

(The template has no "performance" box; the bulk of the diff is a perf change to the
dual quant, the rest are bug fixes.)

## Changes

Performance -- grouped MXFP4 dual quant (`mxfp4_grouped_quant.py`, `mxfp4_quant_kernel.py`):

- Cast the packed 16-bit source directly. bf16->f32 is exact, so both the amax reduction
  and the fp4 convert run on the packed LDS word instead of per-element extracts. This
  halves the abs+max tree and is the single biggest win: 96 of the 104 VALU ops per
  microblock in the ISA, 50 after. The block max stays exact -- no sampling, no
  reduced-width amax.
- Take the amax on packed 16-bit pairs via a `v_perm_b32` selector plus a paired-abs
  mask, halving VALU issue again.
- BM=256 tiles with a c-major `ldsc` staging buffer, so the colwise (wgrad) operand is
  written back in coalesced vec4 stores instead of a transposed scatter.
- Merge the four output words into one b128 store; stream the two write-once fp4 streams
  (410 MB, no reuse) with a non-temporal cache modifier so they leave L2 to the scale
  streams, which cannot reach DRAM as whole sectors any other way.
- Skip the global read, the LDS staging, both barriers and both cast phases on
  all-padding tiles; they still write their zeros, which the wgrad contracts over.
- Adopt the 256-row per-group colwise contract from #468 (producer, consumer, and the
  runtime `nval` bound), replacing the 512-row one.

Correctness -- MXFP4 wgrad whole-loop (`gemm_mxfp4_kernel.py`, `grouped_gemm_mxfp4_kernel.py`):

- Drop the half-M tile fold. It broke bit-exactness on the boundary tile whenever
  `OUT_M % BLOCK_M <= BLOCK_M/2` (e.g. N=2880, last 256-block only 64 rows live), which
  showed up as ~25% cold-run failures of the deterministic test. Net -46 lines, no
  measurable cost.
- Pick the phase-barrier lgkmcnt budget per schedule instead of globally. The budget is
  calibrated against how many `ds_read`s the body leaves in flight, and the fused-C-store
  schedule (bf16) reorders them enough that a non-zero budget lets the next
  `buffer_load_lds` overwrite LDS under them; the standalone-store schedule (fp16) needs
  the 9 it was tuned with, and draining it there costs correctness.
- Make the runtime `nval` dispatch correct for small counts. Under the 512 contract a
  group was always an even number of 256-blocks, so the odd and zero entries were
  unreachable; the 256 contract reaches them. Fixes a parity scratch register aliasing
  the hardware loop bound, three sets of hard-coded local labels colliding across the
  three emission sites, a missing accumulator materialization on the empty-group path,
  and a missing head phase on the odd-and-greater-than-one path.
- Drop the stale scale reload from the runtime odd tail. It had no reader (the tail runs
  on set 0, the reload targets the other half of the ping-pong), and it was still in
  flight when the asm block ended, so for a `wg_tiles>1` workgroup the straggler landed
  on top of the next tile's own scale registers -- that tile came out with `inf` in C,
  intermittently, since it races the next tile's prologue. Set 0 already holds the
  block's scales on both entries: the nval=1 entry still carries the prologue's set, and
  the loop runs an even phase count here, so the ring lands back on set 0 with its last
  prefetch aimed at this block. The compile-time odd tail has always relied on exactly
  that and issues no reload either.

Helpers (`gemm_helper.py`): add `current_stream`, `xcd_remap_pid_blocked`.

## Performance

gfx950 (MI355X), gpt-oss-20b MoE step = quant(act) + fwd + dgrad + wgrad for gate_up and
down; EP1, G=32, per-expert M=4096, hidden 2880 -> 2944, gate_up N=5760, down N=2944.
Same node, same GPU, A/B alternated, FlyDSL + comgr caches dropped before every run.

| | quant | vs #460 | e2e | vs #460 |
|---|---|---|---|---|
| #460 (`05f0f4f7`, merge base) | 0.902 ms | 1.00x | 5.644 ms | 1.00x |
| #468 (`a6a16cdc`, current main) | 0.679 ms | 1.33x | 5.451 ms | 1.036x |
| this PR | **0.449 ms** | **2.01x** | **5.171 ms** | **1.091x** |

Against #468: quant **1.51x**, e2e **1.054x**. The GEMM half is unchanged (4.77 -> 4.72 ms,
within noise) -- the entire gain is the quant.

Skew is a real deployment state for gpt-oss (aux_loss_coeff 0, no expert bias), so the
unbalanced distributions are tracked as guardrails, not scored. All three improve by the
same margin over #468: moderate 5.38 -> 5.14, heavy 5.43 -> 5.20, extreme 5.33 -> 5.11 ms.

Quant SNR against a bf16 dequant reference is 19.0 dB throughout, unchanged from #460 --
the quantization arithmetic is bit-for-bit the same recipe, only the schedule moved.

# Checklist:

- [ ] I have made corresponding changes to the documentation (none needed -- no public API change)

On tests: no new test was needed, the existing coverage already discriminates. Without the
last commit, `test_grouped_gemm_fp4.py -k "mx_blockwise and dtype1 and not deterministic"`
fails 57 / 56 / 59 of 360 across three cache-cold runs (`b_grad_snr=-inf`); with it, 0.
Full results with the fix:

- `test_grouped_gemm_fp4.py` -- 2112 passed
- `test_grouped_gemm_fp4.py --deterministic-only` -- 128 passed, twice
- `test_gemm_fp4.py` -- 181 passed
- `test_quantization.py -k "fp4 or mx"` -- 2305 passed

### DIFF
--- a/primus_turbo/flydsl/gemm/gemm_mxfp4_kernel.py
+++ b/primus_turbo/flydsl/gemm/gemm_mxfp4_kernel.py
@@ -552,6 +552,19 @@ class MfmaScaleFp4:
         _NOD = len(split[0]) if _SPLIT else 0
         if _SPLIT:
             assert n_sub == 2 and nbuf == 2 and nbuf_b == 2 and nsa % 2 == 0 and nsb % 2 == 0
+        # K-loop shape: half_k + even trip count peels the final unroll-2 iteration (the loop stops
+        # two phases early); an odd trip count peels its phase A instead; an odd trailing 128-K
+        # block with no peel runs as an MFMA-only tail.
+        _KPEEL = half_k and not _COOP and (ki is not None) and (ki >= 4) and not (ki & 1)
+        _OPEEL = _CST and half_k and not _COOP and (ki is not None) and (ki >= 5) and bool(ki & 1)
+        _has_loop = (ki is None) or (ki >= 2)
+        _has_tail = (ki is not None) and bool(ki & 1) and not _OPEEL
+        # Accumulator init folded into the first MFMA's src2 immediate 0: the head phase is peeled
+        # so every quad's opening MFMA writes instead of accumulating, which drops the 256-dword
+        # AGPR pre-clear (one VALU per accumulator dword) from every tile's prologue.  Needs the
+        # accumulators pinned to named AGPRs (_CST); a runtime trip count that skips the head
+        # keeps an explicit clear on that branch alone (emit_acc_clear).
+        _ZACC = _CST and (_has_loop or _has_tail)
         if key not in _cache:
             o_acc = list(range(NT))
             t_a = NT
@@ -590,6 +603,9 @@ class MfmaScaleFp4:
             nout += _NCDV
             o_npv = nout  # runtime peel: hw-loop bound = trip count - 2
             if _RTPEEL:
+                nout += 1
+            o_par = nout  # runtime dispatch: parity scratch, never the loop bound
+            if _RUNTIME:
                 nout += 1
             # scale temp accessors (group: 0=A-g0, 1=A-g1, 2=BL, 3=BR; slot=grp*n_sub+s)
             # _scb[0] = ping-pong scale-set base (0 or nsct), set per phase.
@@ -889,7 +905,7 @@ class MfmaScaleFp4:
                 def flush(self):
                     return self.emit(None, 1 << 30)
 
-            def emit_inplace(nxt_buf, g2sl, half=False, drop_s=False, refill=True, cstq=None):
+            def emit_inplace(nxt_buf, g2sl, half=False, drop_s=False, refill=True, cstq=None, zero=False):
                 # NEXT-K in-place refill, blocked-diagonal (4 A-rows x 8 N-cols); GAVOID: g2s in no-refill slots.
                 bm, bn = 4, 8
                 ncol = 2 * ntb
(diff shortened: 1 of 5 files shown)


RuibinCheung (approved):
LGTM

PR #475, opened 2026-08-24
perf(flydsl): promote quant-only optimization stack

zhitwang17:
## Summary

This PR rebases the grouped MXFP4 Quant optimization work onto `main@220ead50` and combines the validated GEAK Quant kernel stack. 

The code change includes code change from both mine and Chushi Chen. Thanks to Chushi for the optimization discovery.

## Changes

The grouped-Quant dataflow rewrite, metadata prologue, memory-access improvements, float-pipe amax and per-shape tiling  were ported from Chushi's GEAK optimization result. 
The buffer-reuse cache and XCD-aware mapping come from the original work and was adapted to the updated kernel layouts and metadata scratch.

- Rework grouped Quant so row-wise and column-wise work can execute concurrently.
- Coalesce/vectorize the dominant output and scale stores.
- Hoist repeated per-group routing scans into a compact metadata prologue.
- Reduce redundant column reads and LDS bank conflicts.
- Fold RHT scaling into scale generation and use the float pipeline for amax.
- Select weight-Quant tile geometry per GPT-OSS shape.
- Retain the validated 8-XCD-aware `_dual3` workgroup mapping.
- Preserve the current-main 256-aligned grouped-Quant/wgrad producer-consumer contract.
- Reuse row/column outputs, scales, padded lens/offsets, and metadata scratch through a bounded 12-entry cache.
- Keep experimental producer-consumer scale-layout fusion opt-in until matching GEMM consumer changes are validated and promoted atomically.

## Performance

Measured on MI355X/gfx950 against `main@220ead50` with GPT-OSS-20B EP1/G32, 24 production routing representatives, GG1 `(N=5760, K=2880)` and GG2 `(N=2880, K=2880)`, using three independent ABBA/BAAB paired sessions.

| Metric | Speedup vs main | Change |
|---|---:|---:|
| Quant | **1.382245x** | **+38.224%** |
| Forward | 1.006247x | +0.625% |
| Dgrad | 1.002133x | +0.213% |
| Wgrad | 0.997759x | -0.224% |
| Backward | 0.999863x | -0.014% |
| Quant + fwd + dgrad + wgrad | **1.104610x** | **+10.461%** |
| Direct full-op | **1.102551x** | **+10.255%** |

## Validation

- Full-op correctness: **48/48 PASS**
- Zero fallback: **PASS**; fallback count: **0**
- Minimum sampled SNR: fwd `14.782 dB`, dgrad `14.779 dB`, wgrad `13.418 dB`
- Python compilation and Git whitespace checks: **PASS**
- The branch defaults to the same producer/consumer fusion settings used by the reported validation; no external environment overrides are required.

## Known limitations

- The shared host showed periodic launch-gap outliers, so the strict same-session A/A MDE promotion gate did not pass. Nevertheless, main-to-combined Quant was `1.382105–1.384005x` and direct full-op was `1.101660–1.103430x` across the three independent sessions.
- Buffer reuse trades memory for lower allocation overhead. The 12-entry bound can retain roughly 9.5 GiB for the tested E1/G32 geometry.
- Re-entering grouped Quant with the same input pointer while retaining an earlier output, or concurrently using that input across streams, may overwrite cached output storage. A production follow-up should use an explicit workspace or stream-aware pool and retain a runtime disable switch.

### DIFF
--- a/primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
+++ b/primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
@@ -1,10 +1,16 @@
 ###############################################################################
+# SPDX-License-Identifier: Apache-2.0
+#
 # Copyright (c) 2025, Advanced Micro Devices, Inc. All rights reserved.
 # Copyright (c) 2025 FlyDSL Project Contributors
-# Adapted from FlyDSL (https://github.com/ROCm/FlyDSL); see LICENSE-APACHE for the Apache-2.0 terms.
 #
-# See LICENSE for license information.
+# Adapted from FlyDSL (https://github.com/ROCm/FlyDSL)
+# Modified by the Primus-Turbo team.
+#
+# This file is distributed under the Apache License 2.0 (see LICENSE-APACHE),
+# not the MIT license that covers the rest of Primus-Turbo (see LICENSE).
 ###############################################################################
+
 """Fused grouped MXFP4 dual-cast quant (rowwise tight-M + colwise 256-aligned-M).
 
 Drop-in for the HIP ``grouped_quantize_mxfp4_dual`` (non-shuffle, per-1x32 E8M0).
@@ -35,9 +41,9 @@ from primus_turbo.flydsl.quantization.mxfp4_quant_kernel import (
     _cvt_microblock_to_fp4,
     _microblock_amax_f,
     _microblock_vf,
-    vf_exp_up,
     _next_sr_seed,
     _sr_hash,
+    vf_exp_up,
 )
 from primus_turbo.flydsl.utils.gemm_helper import (
     _readfirstlane_i32,
@@ -46,6 +52,7 @@ from primus_turbo.flydsl.utils.gemm_helper import (
 )
 
 MB = 32  # MXFP4 microblock (elems per E8M0)
+
 
 # ---- fused fwd/dgrad A-slab scale preshuffle geometry (round 2n, lane `df`) --------
 # The grouped NT GEMM (fwd + dgrad) reads its A-operand E8M0 out of a PACKED, group-
@@ -91,6 +98,7 @@ def rowsc_bytes(total_M, N, G, bm=256, bk=128):
         return total_M * (((N + 127) // 128) * 128 // 32)
     slab_rows, _K256, K128, _KK = rowsc_slab_geom(total_M, N, G)
     return slab_rows * K128 * 4
+
 
 _OOB = 0x7FFFFFFF
 
@@ -323,9 +331,7 @@ def compile_grouped_mxfp4_qdual(
         return (p & fx.Int32(~(_TCW - 1))) + ((p + key) & fx.Int32(_TCW - 1))
 
     # col-phase thread remap: needs the shipped 256-row / 128-feature tile geometry
-    _CMAP = bool(
-        _QZ_CMAP and _SCFUSE and _RMB == 8 and HALF == 256 and _NCOLT == 2 and (BK // 2) == 64
-    )
+    _CMAP = bool(_QZ_CMAP and _SCFUSE and _RMB == 8 and HALF == 256 and _NCOLT == 2 and (BK // 2) == 64)
 
     def _ldsc_swz(q):
         """ldsc word index -> physical, rotated by 4*(cw%8) within each 32-word block.
@@ -337,9 +343,7 @@ def compile_grouped_mxfp4_qdual(
         spread on both the staging write and the linear epilogue read."""
         if not _CMAP:
             return q
-        return (q & fx.Int32(~31)) + (
-            (q + fx.Int32(4) * ((q >> fx.Int32(6)) & fx.Int32(7))) & fx.Int32(31)
-        )
+        return (q & fx.Int32(~31)) + ((q + fx.Int32(4) * ((q >> fx.Int32(6)) & fx.Int32(7))) & fx.Int32(31))
 
     @fx.struct
     class Smem:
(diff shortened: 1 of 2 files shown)


primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
```diff
@@ -556,6 +557,33 @@ def launch(
 
 _GQ_MXFP4_CACHE: dict = {}
 _GQ_MXFP4_CACHE_CAP = 64  # bound the per-(total_M) compiled-quant cache (broad-sweep OOM guard)
+_GQ_REUSE_BUF = os.environ.get("PT_GQ_REUSE_BUF", "1") != "0"
```
RuibinCheung: Please remove debug env.

primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
```diff
@@ -556,6 +557,33 @@ def launch(
...
+_GQ_BUF_CACHE: dict = {}
+_GQ_BUF_CACHE_CAP = int(os.environ.get("PT_GQ_BUF_CAP", "12"))
```
RuibinCheung: same above.

RuibinCheung (changes requested):

primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
zhitwang17: Thanks for the feedback~ These two environment variables are not debug flags; they control the output-buffer reuse optimization and its memory cap. We use `os.environ` because the current operator API does not provide an output/workspace configuration interface.
Our latest ablation measurement shows that disabling reuse causes approximately 1.0% Quant performance loss and 0.13% full-op loss (point estimates). Given the Quant benefit, would it be acceptable to keep and document these environment variables for now?

RuibinCheung: We should be very cautious about adding new environment variables unless they are absolutely necessary. If reuse buf has performance gain. You can set it as default.

kyle-256:
I think our quantize kernel has done with weight-cache buffers, no other cache should be added, since the activation always changes, it's not useful to cache other buffers. I suggest remove this  commit: fe269f24c2b222da198a3b358bc1eba73641937e. @RuibinCheung Please also review it while you can.

primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
```diff
@@ -117,429 +94,354 @@ def _half_to_f32bits(raw16, is_fp16):
...
-    # (col_sc_n = M_pad_col/32, and the fp4 word width 4*col_sc_n = M_pad_col/8) rather
-    # than as baked constants; NBM likewise only exists host-side, in the grid.
+    COL_SC_N = M_pad_col // 32  # colwise scale cols
+    ROW_OUT_W = N_pad // 8  # rowwise fp4 i32 words per row
+
+    _MW = 2  # int32 slots per META entry: (in_rebase, in_end)
+    COL_OUT_W = M_pad_col // 8  # colwise fp4 i32 words per col
+
+    # ---- LDS tile bank swizzle -------------------------------------------------
+    # Measured on the champion (artifacts/PROFILE.md): SQ_LDS_BANK_CONFLICT was 74.7% of
```
kyle-256: remove this debug info

primus_turbo/flydsl/quantization/mxfp4_grouped_quant.py
```diff
@@ -117,429 +94,354 @@ def _half_to_f32bits(raw16, is_fp16):
...
+    def pre(
+        GO: fx.Tensor,  # tight per-group offs (int32 view of int64 [G+1])
+        LC: fx.Tensor,  # OUT: 512-aligned per-group lens (int64 [G])
+        OC: fx.Tensor,  # OUT: 512-aligned per-group offs (int64 [G+1])
+        META: fx.Tensor,  # OUT: int32 [NBM, 2] = (in_rebase, in_end) per padded-M block
+    ):
+        """Per-padded-M-block group metadata, computed ONCE (NBM = 576 threads) instead of
+        once per workgroup x wave inside the main kernel.
+
+        Hardware counters on C8 (artifacts/PROFILE_C8.md): the inline O(G) 512-aligned
```
kyle-256: remove this debug comment

primus_turbo/flydsl/quantization/mxfp4_quant_kernel.py
```diff
@@ -190,64 +248,41 @@ def _microblock_amax(vf):
...
+    vf = _microblock_vf(vbits, use_rht, fold_scale=True)
+    amax = _microblock_amax_f(vf)
+    native_bits, biased = _compute_scale_native(amax, exp_up=vf_exp_up(use_rht))
     words = _cvt_microblock_to_fp4(vf, arith.bitcast(T.f32, native_bits), seed)
     return words, biased
 
 
 # ---- fused-dual tile geometry (shared by the 2D and batched-3D kernels) ----
-_TR = 64  # tile rows (R dim); 2 col-microblocks/tile, 32KB LDS (occ 2)
+# Tile rows/cols for the fused dual. _TR drives the COL_OUT write granule
```
kyle-256: remove this debug comment. no "5760"/ "2880" should be added into the comment.

zhitwang17:
> I think our quantize kernel has done with weight-cache buffers, no other cache should be added,

Thanks for the suggestion! It makes sense since the activation changes frequently and the buffer might lead to data conflict during multi-stream, I will check the actual benefit from the buffer commit and remove it if there is no obvious regression.

zhitwang17:
> > I think our quantize kernel has done with weight-cache buffers, no other cache should be added,
> 
> Thanks for the suggestion! It makes sense since the activation changes frequently and the buffer might lead to data conflict during multi-stream, I will check the actual benefit from the buffer commit and remove it if there is no obvious regression.

@kyle-256 The ablation shows that removing the output-buffer cache increases Quant-only latency by a median of **1.67%** and direct full-op latency by **0.45%**, which is acceptable from my perspective. 
In return, it avoids retaining about **1.16 GiB** for the two production shapes and removes the cross-call/multi-stream overwrite risk caused by buffer reuse. Given this tradeoff, I have followed your suggestion and remove commit fe269f2. Thanks.

RuibinCheung (approved):
LGTM

PR #477, opened 2026-08-24
fix(grouped-gemm): cap FP8 BLOCK_K at 64 on gfx1250

JohnQinAMD:
Triton miscompiles fp8 tl.dot at certain BLOCK_K on gfx1250 -- reproducible with a standalone ~25-line Triton matmul, no primus_turbo involved, and independent of trans_b. Both FP8 forward config pickers select BLOCK_K=128 for the TN layout, so trans_b=True (the layout PrimusTurboGroupedLinear uses for MoE experts) hits it and silently returns NaN at every output column divisible by 8.

A hardcoded per-arch cap goes stale: Triton 3.6.0 breaks BLOCK_K=128, while 3.7.1 additionally breaks BLOCK_K=64 (different signature, columns not %8) and only 32 survives. So select the tile size by PROBING instead: largest_sane_fp8_block_k() runs a small fp8 tl.dot at the requested BLOCK_K, verifies the output is finite and within 20 dB SNR of an fp32 reference, and steps down a power of two otherwise. lru_cached (one small compile per distinct tile size per process), warns once on step-down, and raises with an actionable message if nothing works.

Ruled out our kernel: dropping the tl.multiple_of / tl.max_contiguous hints and the matrix_instr_nonkdim / kpack knobs all still NaN. hipBLASLt is correct and bf16 tl.dot is clean at every tile size, so this is fp8 codegen specific.

This is a WORKAROUND for a Triton bug and should be revisited once the backend is fixed.

Adds test_grouped_gemm_fp8_tn_no_nan (12 combinations), verified failing before and passing after. tests/conftest.py skips the whole suite on gfx1250, which is why existing coverage never caught this; tests can now opt out with @pytest.mark.gfx1250.

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/primus_turbo/triton/grouped_gemm/grouped_gemm_fp8_kernel.py
+++ b/primus_turbo/triton/grouped_gemm/grouped_gemm_fp8_kernel.py
@@ -91,9 +91,17 @@ class ProbeUnavailable(RuntimeError):
 
 @triton.jit
 def _fp8_block_k_probe_kernel(
-    A, B, C, K, N,
-    BM: tl.constexpr, BN: tl.constexpr, BK: tl.constexpr,
-    TRANS_B: tl.constexpr, CACHE_A: tl.constexpr, CACHE_B: tl.constexpr,
+    A,
+    B,
+    C,
+    K,
+    N,
+    BM: tl.constexpr,
+    BN: tl.constexpr,
+    BK: tl.constexpr,
+    TRANS_B: tl.constexpr,
+    CACHE_A: tl.constexpr,
+    CACHE_B: tl.constexpr,
 ):
     """fp8 matmul mirroring the real kernel's inner loop: same operand layout,
     ``tl.multiple_of`` hints and cache modifiers, so the probe compiles the same
@@ -144,8 +152,17 @@ def _fp8_block_k_is_sane(
         )
         out = torch.empty(M, N, device=device, dtype=torch.float32)
         _fp8_block_k_probe_kernel[(M // block_m, N // block_n)](
-            a, b, out, K, N, BM=block_m, BN=block_n, BK=block_k,
-            TRANS_B=trans_b, CACHE_A=cache_a, CACHE_B=cache_b,
+            a,
+            b,
+            out,
+            K,
+            N,
+            BM=block_m,
+            BN=block_n,
+            BK=block_k,
+            TRANS_B=trans_b,
+            CACHE_A=cache_a,
+            CACHE_B=cache_b,
         )
         torch.cuda.synchronize()
     except Exception as exc:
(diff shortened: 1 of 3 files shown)


HuangWei-95:
## Review: clean arch gating, three issues to fix

Merges cleanly with `main`. The gating is well done and I want to call that out: `_PROBED_ARCHS = frozenset({"gfx1250"})` with an early return means the autotune config tables are untouched, no `BLOCK_K=128` config is removed, and **MI300X / MI355X behaviour is bit-identical**. That is the right shape for this kind of workaround.

Three issues:

### 1. The probe does not mirror the kernel it guards

Commit `70e62b64`'s message says the probe was changed to "mirror the kernel it guards", but the real launch passes `matrix_instr_nonkdim=16, kpack=1, num_warps=8` and the probe passes none of them. Those knobs are exactly the ones this file's own comments identify as the trigger for the fp8 miscompile. As written the probe can be both false-negative (miscompile does not reproduce without the knobs, so an unsafe `BLOCK_K` is accepted) and false-positive (a different failure mode caps `BLOCK_K` unnecessarily). Please pass the same knobs the guarded launch uses.

### 2. The cache key omits operand dtype, and HYBRID needs `BLOCK_K=128`

`_fp8_block_k_is_sane` caches on the arch only, and the probe hard-codes `e4m3 x e4m3`. But lines 200-203 of this same file state that HYBRID (`e5m2 x e4m3`) **requires** `BLOCK_K=128`, because below it the e4m3 operand gets decoded as e5m2 and large codes read back as Inf/NaN.

So if the e4m3 probe decides "fall back to 64", that 64 is then reused for HYBRID dgrad — walking straight into the bug this PR is trying to fix. The cache key needs the operand dtype pair, and HYBRID must be pinned at 128 regardless of the probe result.

### 3. `props.gcnArchName` is read unguarded, on the gfx942/gfx950 path too

`_device_arch()` accesses `props.gcnArchName` directly, and that line executes on gfx942 and gfx950 as well — the early return happens *after* it. The existing convention in this repo is `getattr(props, "gcnArchName", "")`. Simpler still: this file already imports `is_gfx950`, so use `is_gfx1250()` from `core/utils.py` and drop the string handling entirely.

### Scope vs title

The title reads as a general gfx1250 `BLOCK_K` cap, but the cap is applied at only two points (tensorwise and rowwise forward). Blockwise forward, the 16 variable-K autotune configs, and MXFP8 all still hard-code `BLOCK_K=128`. Either widen the fix or narrow the title so the remaining gaps are not mistaken for covered.

### On the `tests/conftest.py` change

This is the part I would like to see kept and reused. `conftest.py:53-68` currently skips **all** tests on gfx1250 unconditionally, so gfx1250-specific tests added elsewhere (e.g. in #499 and #481) never execute on any machine — their own `skipif` skips them off-gfx1250, and this conftest skips them on it. The `gfx1250` marker added here is the only working way out, and the other gfx1250 PRs should adopt it.

PR #484, opened 2026-08-27
fix(flydsl): stop recompiling the mxfp4 grouped quant and wgrad on every token count

kyle-256:
# Description

#473 removed `m_total` from the mxfp8 compile keys. The same defect exists on the mxfp4 side, in
the grouped quant and the variable-K wgrad — but it does **not** take the same fix.

In mxfp8 `m_total` was already a launch argument, so striking it from the key was enough. In mxfp4
the padded-M extent is **baked into the generated code** (grid, SRD record counts, fp4 row stride,
packed-scale row stride). Dropping it from the key alone would return a kernel built for a different
contraction length, off by a stride — no fault, no warning, just wrong results. It has to become a
launch argument first.

Since that extent follows the per-step token count (`M_pad_col = ceil256(total_M + G*256)`), a MoE
run recompiled both kernels every step: ~1.97 s of JIT against a 0.74 ms steady state.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Grouped quant**: `NBM` moves host-side into `grouped_mxfp4_qdual_grid()`; the two colwise output
  widths become launch arguments (`col_out_w = col_sc_n << 2` in-kernel). `total_M` was only ever in
  comments. The key keeps just what the kernel still bakes.
- **Removed the 64-entry cap on `_GQ_MXFP4_CACHE`** — a band-aid that cleared the whole dict instead
  of evicting one, making the thrash worse than no cap.
- **Wgrad**: `M_total` becomes an `fx.Int32` argument. It only fed addressing, never the grid, so the
  kernel derives `m2` and `k128m` from it and threads those into `fp4_g2s_offsets`,
  `ScaleS2RPacked`, the rebased int64 SRDs, the `br` LDS-half shift and the scale soffsets — all
  already runtime-capable, as the mxfp8 wgrad shows. Preshuffle grid follows.
- **Wgrad workspace** keyed on `K128m`, stranding a slab pair per length; now keys on the static
  shape and grows in place, like the NT path.
- **`xcd_remap_pid_blocked_dyn()`**: the existing remap branches on `full == 0` / `full ==
  total_pids` in Python, valid only while `total_pids` folds to a constant. The new variant always
  emits the select. Separate entry point so constant callers keep byte-identical ISA, per
  `xcd_remap_pid_u`.
- **Corrected the NT AT-cache comment**, which advertised a `total_M` the key never contained. The NT
  path itself was already M-generic and is unchanged.
- The tile-config selectors keep `M_total` on purpose: thresholds over 2 and 3 fixed configs, already
  in the launch key, so bounded compiles rather than one per token count.

## Validation

MI355X, gpt-oss-20b MoE (`G=32`, `N=2880`, `OUT_M=OUT_N=2880`), token count jittering around 131072
over 8 steps, cleared FlyDSL cache per arm.

**Compiled kernels after 7 distinct `total_M`** — quant / wgrad-launch / wgrad-artifact /
wgrad-workspace: **7,7,7,7 → 1,1,1,1**.

**First call at a new token count**: 1.97 s before (0.22 quant + 1.74 wgrad) vs 0.74 ms after — the
cache-hit cost. Over the 8 steps: **14.09 s → 2.23 s**.

**Bit-exact**: all 5 outputs byte-identical to the old kernels at all 7 token counts (35/35).

**Steady state** (20 iters, `total_M=131072`): quant 0.232 → 0.233 ms, wgrad 0.509 → 0.514 ms (min);
medians move −0.5% / +0.4%. Within noise both directions.

# Checklist:

Unticked deliberately: validation came from an out-of-tree A/B harness, not a committed test — a
regression test asserting `len(_GQ_MXFP4_CACHE) == 1` across several `total_M` would be worth having,
happy to add it here. No docs: internal kernel plumbing, no user-facing surface.

Not covered: I did not diff VGPR counts. The flat steady-state timings argue against new spill; if
you want it hard, a `FLYDSL_DUMP_IR=1` register-count comparison is the check to ask for.

### DIFF
--- a/primus_turbo/flydsl/grouped_gemm/grouped_gemm_mxfp4_kernel.py
+++ b/primus_turbo/flydsl/grouped_gemm/grouped_gemm_mxfp4_kernel.py
@@ -542,7 +542,7 @@ def _build_grouped_mxfp4_nt_kernel(
 
 _GMXFP4_LAUNCH_CACHE: dict = {}
 _GMXFP4_WS_CACHE: dict = {}
-_GMXFP4_AT_CACHE: dict = {}  # (total_M, N, K, G, gm, xcd, gn, out_fp16) -> [raw_launch, compiled]
+_GMXFP4_AT_CACHE: dict = {}  # (N, K, G, gm, xcd, gn, span, nt, out_fp16, k_real) -> [raw, compiled]
 _GMXFP4_NT_CFG = (4, 8, 0, 16, False)
 # Thin groups: the write-only C stream evicts re-read weights, so non-temporal buys them back.
 _GMXFP4_NT_CFG_THIN = (4, 8, 0, 16, True)
@@ -699,7 +699,6 @@ def _build_grouped_mxfp4_wgrad_kernel(
     OUT_M,
     OUT_N,
     G,
-    M_total,
     group_m=4,
     num_xcds=8,
     group_n=0,
@@ -716,7 +715,6 @@ def _build_grouped_mxfp4_wgrad_kernel(
     N_SUB = BLOCK_K // 128
     BPR = BLOCK_K // 2
     KSTEP = BPR
-    M2 = M_total // 2  # operand row stride (bytes) = full contraction
     N_TILES_A = BLOCK_M // 32
     LDS_BN_HALF = BLOCK_N // 2
     N_TILES_BH = LDS_BN_HALF // 32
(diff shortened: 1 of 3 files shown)


RuibinCheung (dismissed):
LGTM

RuibinCheung (approved):
LGTM

PR #490, opened 2026-09-03
perf(fp8): grouped GEMM kernel opts for gpt-oss-20b padded MoE

suachong:
## Summary
- FlyDSL tensorwise FP8 grouped GEMM campaign on padded gpt-oss-20b shapes (K 2880→2944), rebased onto current `main` (includes #488 fused MLP pad).
- NT: folded-N epilogue (`compute_global_swizzle_fold` + `StoreCPerTensorQuadN`) so a lane's two N-fragments store as one dword.
- NN: 64-col narrow last-N body (`nn_bnd`) plus row-merge `StoreCPerTensorRowN` when columns pair cleanly; g2s/vmcnt/barriers kept identical to the existing half body.
- TN wgrad: static affine band-width window `_WGRAD_AFF_W=(3,4)` so autotune does not pick a band that overflows an XCD's 4 MiB L2 slice; last-M boundary body kept when last-N already runs.

Campaign ruler (balanced, six shapes, SNR ≥ 25 dB + bitwise determinism): **+2.25%** score vs `220ead50`. Synthetic skew A/B (`skew_3to1` / `ramp` / `powerlaw` / `half_empty` / `one_hot_dust`) stayed **+1.2–1.9%** vs the same base, all `ok=true`.

## Test plan
- [ ] `CAMPAIGN_PAD=1` six-shape `bench_gptoss_fp8_grouped.py` vs `main`
- [ ] Skew presets above (correctness + no score flip)
- [ ] Smoke fused `grouped_mlp_fp8` / e2e on this branch (campaign timed the kernel directly, not the #488 wrapper)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/primus_turbo/flydsl/grouped_gemm/grouped_gemm_fp8_kernel.py
+++ b/primus_turbo/flydsl/grouped_gemm/grouped_gemm_fp8_kernel.py
@@ -50,6 +50,7 @@ from primus_turbo.flydsl.utils.gemm_helper import (
     S2RLoaderTr,
     StoreCPerTensor,
     StoreCPerTensorCShuffle,
+    StoreCPerTensorQuadN,
     StoreCPerTensorRowN,
     _lane_tbl_count_le,
     _lane_tbl_get,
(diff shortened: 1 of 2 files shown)


RuibinCheung (approved):
LGTM

PR #492, opened 2026-09-03
perf(flydsl): tune the dense per-tensor FP8 GEMM whole-loop on gfx950

kyle-256:
# Description

Performance tuning of the FlyDSL backend of the dense per-tensor FP8 GEMM
(`gemm_fp8_tensorwise_flydsl_kernel`) on gfx950/MI355X.

The thin-N (3072) deep-K shapes of the gpt-oss-20b MLP carry the bulk of dense FP8 GEMM time,
and all three layouts sit within a couple of percent of `torch._scaled_mm` there. This PR works
that regime: the mfma issue order becomes a per-geometry property, the TN tile gets a
whole-line store, the NT/NN whole loop gains a narrower macro tile for shapes the square one
cannot round onto whole CU passes, and two dispatch-side allocations leave the steady-state
launch path.

Kernel behaviour is unchanged: same bf16 -> fp8 e4m3 cast, same fp32 accumulate on the f8f6f4
mfma, same quantiser, same scale semantics. No API change.

Fixes # (issue) — N/A, no tracking issue.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

(Performance change; the closest fit is "new feature" since it adds a macro tile and a store
path to the autotune candidate set.)

## Changes

- Make the mfma issue order of the wave4 tiles a geometry field instead of a constant. Both
  operands of the square TN tile come through the transpose reader, whose four reads per
  fragment issue cheaper as one burst than spread over a diagonal, so that tile takes a
  whole-row column step and a four-row row step; the transposed-B NT variant takes the same
  step on its B side only.
- Give the TN tile the whole-line store. Both its operands arrive transposed, so the epilogue
  is the one place it can shed issue slots: the same bytes go out in the same 64 B requests in
  half the store instructions (128 -> 64 per wave per tile), and the per-tensor scale is
  hoisted out of the tile with it.
- Add a narrower 256x192 macro tile to the NT/NN whole loop for shapes whose tile count lands
  between two whole CU passes, where the square tile leaves a quarter of the device unwoken.
  It joins the first-call race only where the coverage it wakes outweighs the operand bytes it
  pulls (`_nt4_geoms`).
- Build the NT/NN store per n-extent rather than per tile, so the narrow tile's uneven B side
  is supported; an uneven cut keeps the plain store throughout.
- Carry the NT whole loop's pool fills across the tile boundary, as NN already did.
- Pin the narrow tile's XCD group instead of racing it — the two arms are too far apart on
  deep K for a first-call race to rank reliably.
- Replace the TN plan's makespan model with the CU passes each macro tile occupies. This only
  orders the race; the race still picks. (Ranking is not a substitute for measuring: taking the
  plan's first entry unmeasured costs up to 14% on the llama-2 wgrad shapes.)
- Return a tuned shape from the cache before its dispatch assembles a candidate list, and skip
  each no-op step in `_scalar_scale` instead of only the all-no-op case — both remove
  allocations from the steady-state launch path.

## Results

FlyDSL vs frozen FlyDSL (`ed8d7af4`), STREAM-timed on GPU3/MI355X, median of 3 alternating
paired runs. `share` is each shape's fraction of production FP8 GEMM time.

| cell | M / N / K | share | main ms | this PR ms | speedup |
|---|---|---:|---:|---:|---:|
| NT / qkv_d | 8192 / 9216 / 3072 | 3.52% | 0.1950 | 0.1888 | 1.0328 |
| NT / proj_d | 8192 / 3072 / 3072 | 1.36% | 0.0772 | 0.0705 | **1.0950** |
| NT / mlpup_d | 8192 / 12288 / 3072 | 4.06% | 0.2314 | 0.2266 | 1.0212 |
| NT / mlpdn_d *(P0)* | 8192 / 3072 / 12288 | 3.69% | 0.2465 | 0.2195 | **1.1230** |
| NT / lin1_s | 16384 / 21504 / 3072 | 13.48% | 0.7565 | 0.7336 | 1.0312 |
| NT / lin2_s *(P0)* | 16384 / 3072 / 15360 | 8.24% | 0.4888 | 0.4883 | 1.0010 |
| NN / qkv_d | 8192 / 3072 / 9216 | 3.01% | 0.1935 | 0.1838 | 1.0528 |
| NN / proj_d | 8192 / 3072 / 3072 | 1.25% | 0.0764 | 0.0738 | 1.0352 |
| NN / mlpup_d *(P0)* | 8192 / 3072 / 12288 | 3.59% | 0.2456 | 0.2317 | 1.0600 |
| NN / mlpdn_d | 8192 / 12288 / 3072 | 4.01% | 0.2210 | 0.2207 | 1.0014 |
| NN / lin1_s *(P0)* | 16384 / 3072 / 21504 | 11.63% | 0.6683 | 0.6672 | 1.0016 |
| NN / lin2_s | 16384 / 15360 / 3072 | 9.58% | 0.5499 | 0.5528 | 0.9948 |
| TN / qkv_d *(not FP8 in production)* | 9216 / 3072 / 8192 | — | 0.2013 | 0.1863 | 1.0805 |
| TN / proj_d | 3072 / 3072 / 8192 | 1.08% | 0.0801 | 0.0786 | 1.0191 |
| TN / mlpup_d | 12288 / 3072 / 8192 | 3.71% | 0.2228 | 0.2242 | 0.9938 |
| TN / mlpdn_d | 3072 / 12288 / 8192 | 3.82% | 0.2190 | 0.2190 | 1.0000 |
| TN / lin1_s *(P0)* | 21504 / 3072 / 16384 | 12.06% | 0.6926 | 0.6824 | 1.0149 |
| TN / lin2_s | 3072 / 15360 / 16384 | 11.91% | 0.5125 | 0.5039 | 1.0171 |

Weighted by share:

| set | cells | share | speedup |
|---|---:|---:|---:|
| **full deployment table** | 18 | 100% | **1.0158** |
| P0 (thin-N / deep-K) | 5 | 39.2% | 1.0134 |
| P1 / P2 | 12 | 60.8% | 1.0177 |

No cell regresses (none below 0.99); the two lowest, NN/lin2_s at 0.9948 and TN/mlpup_d at
0.9938, sit inside this ruler's ~0.3% run-to-run spread. The two shapes previously flagged as
must-not-regress both improve: TN 3072x15360x16384 to 1.0171 and NT 8192x3072x3072 to 1.0950.
The largest single share, NT 16384x21504x3072 (13.48%), gains 3.1%.

Sibling shapes outside the tuned regime — llama-2 7B/70B, m=4096/8192, all three layouts,
48 cells — come out at 1.0079 overall with no wgrad cell below parity.

Correctness: FlyDSL vs an fp32-dequant reference, one case per layout, 55.6 / 55.6 / 52.5 dB
against a 45 dB floor.

# Checklist:

- [ ] I have made corresponding changes to the documentation — N/A, no user-facing docs cover
      this backend's internals

### DIFF
--- a/primus_turbo/flydsl/gemm/gemm_fp8_kernel.py
+++ b/primus_turbo/flydsl/gemm/gemm_fp8_kernel.py
@@ -1123,6 +1123,8 @@ class _Tn4Geom(NamedTuple):
     bn: int
     pools: tuple
     bstep: int
+    bstep_tr: int = 0  # bstep for the transposed-B variant, whose reads issue best as one burst
+    mstep: int = 2  # srcA rows of one issue-order step, i.e. how long a srcB fragment is held
     # Boundaries taking a wait state, "<before><after>" over m (mfma), d (ds_read), g (g2s).
     pad: tuple = ("mm", "md", "mg", "dm", "dd", "dg", "gm", "gd", "gg")
     wgrid: tuple = (2, 2)  # a square split shares both operands' reads four ways, not one side's
@@ -1141,7 +1143,8 @@ _TN4_SQUARE = _Tn4Geom(
     256,
     256,
     ((0, 128, 2), (0, 128, 2), (1, 128, 3), (1, 128, 3)),
-    0,
+    8,  # both operands read through the transpose reader, whose reads issue best as one burst
+    mstep=4,
 )
 # Finer grid for shapes the square tile cannot round onto whole CU passes, same operand bytes.
 _TN4_RECT = _Tn4Geom(
@@ -1695,8 +1698,23 @@ def _dense_tn_wave4_tile(
     mfma = {p.tiles: Mfma16x16x128(apool[0].tiles, p.tiles) for p in bpool}
     if c_rows is None:
         c_rows = fx.Int32(M) if row_shift is None else fx.Int32(M) + row_shift
+    pool_lds = [getattr(lds, f"p{i}") for i in range(len(pools))]
+    # Both operands arrive transposed, so the epilogue is where this tile sheds issue slots.
+    line = _nt4_line_extents(0, col_safe, tuple(mfma))
+    line_kw = (
+        {
+            "lds_xpose": _nt4_xpose_lds(pools[0], pool_lds[0], wave_id, _tn4_nthr(geom) // 64),
+            "scale": load_per_tensor_scale(A_scale, B_scale),
+        }
+        if line
+        else {}
+    )
     store_c = {
-        nb: (StoreCPerTensorPairN if col_safe and nb % 2 == 0 else StoreCPerTensor)(
+        nb: (
+            StoreCPerTensorLineN
+            if nb in line
+            else (StoreCPerTensorPairN if col_safe and nb % 2 == 0 else StoreCPerTensor)
+        )(
             A_scale,
             B_scale,
             C,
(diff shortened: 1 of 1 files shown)


RuibinCheung (approved):
LGTM

PR #494, opened 2026-09-07
feat(quant): support FlyDSL MXFP4 UoS modes

zhitwang17:
## Summary

This PR adds configurable MXFP4 UoS scale-rounding modes to the FlyDSL quantization path.

## Changes

- Add `PRIMUS_TURBO_MXFP4_SCALE_ROUNDING` with three supported modes:
  - `0` or unset: existing half-ULP bias (default)
  - `1`: one-ULP bias
  - `2`: three-eighth-ULP bias
- Reject invalid values instead of silently selecting an unintended rounding mode.
- Apply the selected rounding semantics consistently to both the HIP and FlyDSL implementations.
- Cover the three FlyDSL MXFP4 quantization paths used by GPT-OSS training:
  - Dense dual quant
  - Batched 3D expert dual quant
  - Grouped dual quant
- Pass the bias as a launch-time value for dense and grouped quantization, avoiding unnecessary kernel recompilation. Batched 3D quantization caches at most one compiled variant per mode.
- Preserve the existing RHT, 2D block scaling, stochastic rounding, padding, and grouped-layout behavior.

## Validation

Added seven focused UoS test functions covering:

- Default, supported, and invalid mode parsing
- Dense runtime mode switching
- Byte-exact FlyDSL/HIP parity for RHT and 2D block-scaling recipes
- Padded batched 3D quantization and per-mode compile-cache switching
- HIP rejection of invalid modes
- Grouped quantization with empty and irregular K256 expert spans
- Stochastic-rounding scale parity

The test refactor removes redundant `mode × kernel variant` Cartesian combinations without removing or narrowing any test that exists on `main`.

## Compatibility and Risk

The default mode preserves the current quantization behavior. The environment variable is process-wide and should be configured consistently before launching quantization work. Switching modes for the batched 3D path may retain up to three compiled kernel variants in the process cache.

### DIFF
--- a/csrc/include/primus_turbo/quantization.h
+++ b/csrc/include/primus_turbo/quantization.h
@@ -121,6 +121,21 @@ constexpr int   FP8E4M3_FNUZ_TARGET_MAX_POW2 = 7;
 
 constexpr int E8M0_EXPONENT_BIAS = 127;
 
+inline int mxfp4_scale_rounding_bias(const int64_t mode) {
+    constexpr int shift = FP32_MANTISSA_BITS - FP4_MANTISSA_BITS;
+    switch (mode) {
+    case 0:
+        return 1 << (shift - 1);
+    case 1:
+        return 1 << shift;
+    case 2:
+        return 3 << (shift - 3);
+    default:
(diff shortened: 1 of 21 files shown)


RuibinCheung (approved):
LGTM

PR #498, opened 2026-09-08
fix(flydsl): cover all eligible MXFP4 dense quant rows

zhitwang17:
## Summary

This PR fixes incomplete row coverage in the FlyDSL dense dual MXFP4 quantization kernel. The bug could leave the final rows of the packed FP4 data and E8M0 scales uninitialized, causing NaN/Inf in downstream GEMMs during GPT-OSS-20B training.

## Root Cause

The dense kernel used `_TR = 96`, while its eligibility check accepts every shape with `R % 128 == 0`. The launch grid uses `R // _TR`, so an eligible row count is not always fully covered.

For the production shape `1024 × 4096`, only `10 × 96 = 960` rows were processed. The final 64 rows remained unwritten and could introduce NaN/Inf into downstream GEMMs. The collective timeouts observed after rank failures were a secondary effect.

## Fix

- Restore the dense/fallback row tile from `_TR = 96` to `_TR = 64`.
- Since every eligible row count is divisible by 128, the 64-row tile guarantees complete coverage.
- Keep the optimized batched 3D path unchanged; it can still select 96- or 128-row tiles through `_pick_tile_geom()`.
- Add targeted Quant and Quant-to-GEMM regression tests for row coverage, byte-exact HIP/FlyDSL parity, finite GEMM output, and SNR.

## Validation

Validation on MI355X/gfx950:

- New targeted Quant and GEMM regressions: **2 passed**
- Existing dual-quant parameterized suite: **128 passed**
- FlyDSL reproduction after the fix: **100 consecutive byte-exact runs**, with no unwritten sentinel values or non-finite outputs
- HIPBLASLt and AITER replays: **20 consecutive runs each passed**
- GPT-OSS-20B EP8 training: one complete iteration passed with finite loss and zero NaN iterations
- All **384/384** observed MXFP4 backend calls completed with finite output
- Ruff format, Ruff lint, and `git diff --check`: **passed**

## Performance Regression

Quant-only performance was compared on shapes fully computed by both versions:

| Shape | Before | After | Change |
|---|---:|---:|---:|
| `384 × 4096` | 9.008 μs | 9.026 μs | +0.20% |
| `768 × 4096` | 8.946 μs | 9.045 μs | +1.11% |
| `1152 × 4096` | 8.893 μs | 8.875 μs | -0.20% |
| `1536 × 4096` | 9.062 μs | 8.981 μs | -0.90% |

The aggregate latency change is approximately **+0.05%**, indicating no material performance regression.

The affected `1024 × 4096` case is excluded from the comparison because the old kernel skipped 64 rows, making its previous latency an invalid performance baseline.

### DIFF
--- a/primus_turbo/flydsl/quantization/mxfp4_quant_kernel.py
+++ b/primus_turbo/flydsl/quantization/mxfp4_quant_kernel.py
@@ -277,12 +277,11 @@ def _finish_microblock(vbits, use_rht, seed=None):
 
 
 # ---- fused-dual tile geometry (shared by the 2D and batched-3D kernels) ----
-# Tile rows/cols for the fused dual. _TR drives the COL_OUT write granule
-# (_TR/8 i32 per feature's R-run): at _TR=64 that is 32 bytes, i.e. every byte of
-# the 265 MB colwise weight output leaves in a sub-cacheline burst. _TR must divide
-# both shipped N (5760 and 2880) and _TR*_TC/32 must divide BLK=256.
-# _TC also sets the K tail: ceil(2880/_TC)*_TC.
-_TR = 96  # tile rows (R dim)
+# Tile rows/cols for the fused dual. The 2D eligibility contract admits every
+# R divisible by 128, so its default row tile must divide 128. Batched weight
+# quantization temporarily selects the faster 96/128-row geometry per shape via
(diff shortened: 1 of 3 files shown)


RuibinCheung (approved):
LGTM

PR #499, opened 2026-09-08
feat(gemm): add a gfx1250 FlyDSL GEMM for bf16 / fp8 / mxfp8

JohnQinAMD:
The existing primus_turbo/flydsl/gemm/gemm_bf16_kernel.py does not run on gfx1250: its G2SLoader emits BufferCopyLDS128b (buffer_load ... lds), which this target does not have, its MMA atom is Mfma32x32x16 (gfx1250 has no MFMA at all), and it assumes wave64. CK is not an option on this target either.

This adds the gfx1250 counterpart -- one NT kernel (A[M,K] @ B[N,K]^T) built on TDM descriptor atoms for global->LDS, wave32 WMMA, and a multi-buffer K pipeline.

The legal WMMA atom shapes were probed against the backend rather than taken from documentation. Only the K dimension is narrow:

  bf16 / f16   WMMA        16x16x32 only, and A and B must share a dtype
  fp8          WMMA        16x16x64 / 16x16x128, any e4m3 / e5m2 pairing
  mxfp8        WMMAScale   16x16x128, block_size 16 or 32

16-bit operands ride FlyDSL's tiled-MMA machinery, which derives the per-lane register mapping from the atom. 8-bit operands cannot: allocating a register fragment of an 8-bit type leaves an `ub.poison !fly.ptr<f8E4M3FN, register>` behind after rmem SSA promotion, so those paths carry the operands as packed i32 with an explicit lane mapping, the way the gfx1250 fp8/fp4 kernels in ROCm/aiter do.

Two things that turned out to matter more than the tiling:

* Calling the @flyc.jit entry point directly costs ~69 us of host time per launch (signature rebind + cache-key rebuild). That dominates every shape below roughly 4096^3 -- 1024^3 measured 69 us wall against 8.3 us in a CUDA graph. The wrapper caches a flyc.compile() CompiledFunction instead.
* LDS rows are padded by 16 B. Unpadded, a row is an exact multiple of the 128 B bank stride, so all 16 lanes of a fragment read hit one bank. The TDM descriptor inserts the padding itself (pad_interval / pad_amount), so it costs nothing on the compute side.

Measured on gfx1250 at 4096^3, defaults, no per-shape tuning:

  bf16    1979 TFLOPS   (Triton 1136, torch/hipBLASLt 81)
  fp8     4880 TFLOPS   (aiter a8w8 2127)
  mxfp8   2550 TFLOPS

Licensing: Apache-2.0 (LICENSE-APACHE), not the repo's MIT, matching the other FlyDSL-derived files here. The docstring credits the Apache-2.0 gfx1250 kernels in ROCm/aiter whose TDM/WMMA idiom this follows.

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/primus_turbo/common/constants.py
+++ b/primus_turbo/common/constants.py
@@ -51,6 +51,9 @@ ENV_SPARSE_ATTN_BACKEND = "PRIMUS_TURBO_SPARSE_ATTN_BACKEND"
 # (e.g. "<OP>_BACKEND=HIPBLASLT") still takes precedence over it.
 # Default: "0" (disabled)
 ENV_AUTO_TUNE = "PRIMUS_TURBO_AUTO_TUNE"
+# Make a pinned backend a preference rather than a requirement (see
+# BackendChoice.allow_fallback).
+ENV_BACKEND_ALLOW_FALLBACK = "PRIMUS_TURBO_BACKEND_ALLOW_FALLBACK"
 
 # Whether Attention V3 uses FP32 atomic accumulation ("1" to enable, "0" to disable).
 # Default: "1" (enabled)
--- a/primus_turbo/flydsl/gemm/gemm_gfx1250_kernel.py
+++ b/primus_turbo/flydsl/gemm/gemm_gfx1250_kernel.py
@@ -60,7 +60,9 @@ from flydsl.expr.typing import T as _T
 
 __all__ = [
     "WAVE",
+    "autotune",
     "can_run",
+    "feasible_configs",
     "gemm_gfx1250",
     "launch_gemm_gfx1250",
     "supported_dtypes",
@@ -170,6 +172,7 @@ def launch_gemm_gfx1250(
     out_is_f16: Constexpr[int],
     group_m: Constexpr[int] = 8,
     layout: Constexpr[int] = LAYOUT_NT,
+    accumulate: Constexpr[int] = 0,
 ):
     """NT GEMM: ``C[M, N] = A[M, K] @ B[N, K]^T`` on gfx1250.
 
@@ -212,10 +215,6 @@ def launch_gemm_gfx1250(
     # that path stays byte-identical across all three layouts.
     a_tr = layout == LAYOUT_TN
     b_tr = layout in (LAYOUT_NN, LAYOUT_TN)
-    assert not (a_tr or b_tr) or is_16bit, (
-        "the LDS transpose is wired up for 16-bit operands only so far; "
-        "8-bit NN/TN still needs the ds_load_tr8_b64 path"
-    )
     TR_A_ROW = tile_m + LDS_PAD  # scratch row = the contiguous (M or N) extent
     TR_B_ROW = tile_n + LDS_PAD
     SCRATCH_A = tile_k * TR_A_ROW * elem_bytes if a_tr else 0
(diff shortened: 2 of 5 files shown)


HuangWei-95:
## Review: arch gating is safe, three blocking issues, and please split the dispatcher change out

No conflict with `main`. I checked the compatibility question first and the gating is sound: `can_handle()` returns `False` on its first line when `not is_gfx1250()`, flydsl is imported lazily inside `_flydsl_gemm()`, `BackendType.FLYDSL` already existed on the merge-base (so no `auto()` renumbering of existing enum members), and the new entry is appended last in `_GEMM_BACKENDS` while all three `default_backend` sites still hard-code `HIPBLASLT`. **gfx942 and gfx950 cannot fall into this path.**

Three things I think should block merge:

### 1. `_resolve()`'s LDS precheck can disagree with the kernel it guards

The precheck hard-codes `accumulate=False` (the function does not take the parameter at all), but the kernel sizes its arena as `C_BYTES * (2 if accumulate else 1)`. For bf16 / NT / `M=8192, N=4096, K=4096` with `accumulate=True`, `can_run()` returns `True` and `flyc.compile()` then trips an `AssertionError`. That directly contradicts this function's own docstring, which promises the precheck never diverges from the actual call.

Same function: `ob` is taken from `out_dtype` while the kernel takes it from `_OUT_KIND[out.dtype]`. Megatron passes `out=main_grad` (fp32) with `out_dtype=bf16`, so these differ by 2x on a real call path.

### 2. `can_handle()` must not be able to raise

`setup.py:529-535` skips installing flydsl on gfx1250 specifically. `can_handle()` has no `try/except ImportError`, so on a gfx1250 box without flydsl any ordinary bf16 GEMM that reaches `tune()` or the step-4 dispatch fallback raises `ImportError` — even for a user who never asked for the FLYDSL backend. `can_handle()` is contractually a predicate; it needs to return `False` rather than propagate.

### 3. The dispatcher semantics change is untested and should be its own PR

`BackendChoice.allow_fallback` and `_pinned_may_fall_back` change dispatch behavior for **every operator on every architecture**, but the only test covering them lives in a file guarded by `skipif(not gfx1250)`. Since there is no gfx1250 runner in CI (see below), that means **zero coverage for a global semantic change**. This belongs in the existing `tests/pytorch/core/test_global_backend_manager.py` with a dummy dispatcher, where it is hardware-independent. I would strongly prefer the dispatcher change land as a separate PR from the gfx1250 kernel.

### Naming convention

`CONTRIBUTING.md:50-92` specifies `<op>_<format>_kernel.py` for the FlyDSL layer, where `<format>` is the wire format — its neighbours are `gemm_bf16_kernel.py` and `gemm_fp8_kernel.py`. `gemm_gfx1250_kernel.py` puts an architecture in the dtype slot. Suggest `gemm_bf16_kernel.py` with an arch-specific internal path, or at minimum a name that keeps the format.

### CI note

There is no gfx1250 runner in this repo's CI, and `tests/conftest.py:53-68` applies an unconditional skip to **all** collected tests on gfx1250 and then returns. So the new tests here never execute anywhere: on other hardware their own `skipif` skips them, and on gfx1250 the conftest skips them. #477 adds a `gfx1250` marker to exempt tests from that global skip — that is currently the only working mechanism, and these tests should use it.

PR #500, opened 2026-09-09
Speed up the gfx950 dual-wave forward attention kernel at head_dim 64

wenxie-amd:
Four changes to the FlyDSL forward attention kernel, worth **1.14x on dense** and
**1.38-1.39x on sliding-window** at head_dim 64. Nothing here changes the API or the
numerics, and the LSE is emitted throughout, so the gains apply to training as well as
inference.

### Measurements

AMD Instinct MI355X (`gfx950:sramecc+:xnack-`), bf16, GQA head_dim 64, seq 8192,
`return_lse=True`. Each pair is run in one session, palindromically ordered
(new, base, base, new) so both arms get the same mean position, one process per arm, on
an idle node. Statistic: min over repeats of `mean(Timer.timeit(100))` after 3 s of
continuous warm-up; both arms within 0.2%.

| shape | base TFLOP/s | this branch | speedup | base ms | this branch ms |
| --- | ---: | ---: | ---: | ---: | ---: |
| b3, dense, sink | 1101.1 | **1259.1** | **1.143x** | 1.4980 | 1.3100 |
| b4, dense, sink | 1106.6 | **1261.5** | **1.140x** | 1.9874 | 1.7434 |
| b3, window 128 | 239.3 | **329.2** | **1.376x** | 0.2153 | 0.1566 |
| b4, window 128 | 248.6 | **345.7** | **1.391x** | 0.2764 | 0.1987 |

Taken from a fresh clone of this branch, so the figures are reproducible from what is
here rather than from a working tree.

### Correctness

17/17 cases pass against a torch reference at **50.59-88.37 dB SQNR** on both returned
tensors (floor 40 dB), every case **bitwise reproducible** run to run, including a
`deterministic=True` probe. The set covers ragged and rectangular sequence lengths,
windows of 64/128/192/1024, GQA group sizes 8 and 16, and configurations with and
without a sink.

The emitted code object is checked per shape to be the LSE-emitting variant, rather than
taken on trust from the call arguments — `emit_lse` is a compile-time trait behind an
`lru_cache`, so a stale entry can otherwise serve a variant the caller did not ask for.

### The changes

**1. Defer the attention-sink load to its consumer.** Issued in the prologue, the sink is
live across the whole KV loop for a single consumer in `finalize_o_scale`. That one
long-lived value is what pushes the sink + LSE build over the allocator's limit: it emits
112 B of scratch where every neighbouring configuration emits 0-24 B, and the sink is the
only thing it adds to the loop body. Loading it at the point of use brings that to 28 B.
Same address, same value, same order of fp32 operations. Worth 13.2% on the dense shapes.

**2. Halve `BLOCK_M` under a left window.** A window trims the body to
`ceil((W + block_m)/BLOCK_N)` KV tiles, and that band is `W + block_m` columns wide
however few rows share it — so halving `BLOCK_M` removes a quarter of the masked-away
MFMA work. The narrower body also falls out of the `block_m in (128, 256)` gate that
raises `waves_per_eu` to 4, which costs more than it returns on a body this short. Gated
on the window: the dense path has no band and is bit-for-bit unaffected.

**3. Let the windowed body run an odd number of KV tiles.** The pipeline is prologue +
two-tiles-per-iteration + drain, so the legal trip counts are set by the drain alone. A
128-wide window over a 64-row q block touches exactly three KV tiles; rounding up to even
made every windowed workgroup carry a fourth that the mask then threw away. A two-tile
drain admits the odd counts.

**4. Elide provably dead work at the band edges.** For a windowed band one of the six
half-tiles a wave owns is all `-inf` by construction, and on the dense causal band the
last tiles are wholly past the causal edge. Both predicates are wave-uniform, so these
are scalar branches rather than divergence, and the skipped accumulators are overwritten
by the mask that follows — bit-exact by construction.

### One thing to know before tidying the source

The `const_expr` branch in `_qk_cluster_sync` emits the same instructions either way and
**is not dead code**. Collapsing it to a bare `_dualwave_sync_barrier()` costs **5%**
(1261 -> 1199 TFLOP/s), bisected to that function and confirmed by restoring only that
branch. The tracer appears to make a region of the arms, and the boundary reaches the
scheduler; this kernel is unusually sensitive to that, since sweeping the
`sched_group_barrier` anchor count alone moves it 4%. That is as far as it is understood,
and the comment in the source says so.

### DIFF
--- a/primus_turbo/flydsl/attention/flash_attn_fwd.py
+++ b/primus_turbo/flydsl/attention/flash_attn_fwd.py
@@ -168,6 +168,11 @@ def build_flash_attn_dualwave_swp_module(
         fold_mem_barriers = sink_drains and traits.Q_HEADS_PER_WG > 1
         fold_pv_barriers = fold_mem_barriers and late_dma
         vm_drain_sync = 0 if fold_pv_barriers else ctx.VM_DRAIN_KV
+        # Drain length. The pipeline's legal trip counts are set by
+        # the drain: three named tiles means N even, two means N odd. The windowed body
+        # needs exactly three KV tiles, so it takes the two-tile drain and attn_helper.py
+        # keeps its span odd. Every other configuration keeps the three-tile drain.
+        short_drain = traits.CAUSAL and traits.WINDOW_LEFT >= 0
 
         def _mem_cluster_sync():
             if const_expr(fold_mem_barriers):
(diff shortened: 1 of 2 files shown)


RuibinCheung (approved):
LGTM if CI is green.

PR #501, opened 2026-09-09
fix(quant): carry the opaque metaclass torch 2.11 requires

JohnQinAMD:
register_opaque_type refuses a class whose metaclass is not OpaqueBaseMeta:

  TypeError: Opaque type <class '...Float8QuantConfig'> must subclass
  torch._opaque_base.OpaqueBase or 'metaclass=torch._opaque_base.OpaqueBaseMeta'

The call sits at module scope, so on torch 2.11 this is not one operator failing -- `import primus_turbo.pytorch` raises and nothing in the package can be used. Reproduced on torch 2.11.0+rocm7.14.0a20260625; #483 introduced the registration.

The two dataclasses take the metaclass directly. ScalingRecipe cannot: NamedTuple builds through its own metaclass and Python rejects a metaclass that is not a subclass of every base's, so the fields keep their class (renamed) and a thin derived class carries the metaclass. Tuple behaviour is unchanged -- unpacking, indexing, len, _asdict, _replace, == and hash all still work, and the new test pins that so the split cannot be undone silently.

The metaclass is looked up behind a try/ImportError so torch versions without torch._opaque_base fall back to `type`, which leaves the classes as they were.

The tests are arch-independent, but tests/conftest.py skips everything on gfx1250, so they were verified there with that blanket skip lifted: 11 passed against this change, and collection fails with the TypeError above without it.

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/primus_turbo/pytorch/core/low_precision.py
+++ b/primus_turbo/pytorch/core/low_precision.py
@@ -13,6 +13,13 @@ from torch._library.opaque_object import register_opaque_type
 
 from primus_turbo.pytorch.core.utils import get_device_compute_capability
 
+try:
+    # torch >= 2.11 requires an opaque type to carry this metaclass; older torch has
+    # neither the module nor the requirement, where `type` changes nothing.
+    from torch._opaque_base import OpaqueBaseMeta as _OpaqueMeta
+except ImportError:  # pragma: no cover - depends on the installed torch
+    _OpaqueMeta = type
+
 __all__ = ["float8_e4m3", "float8_e5m2"]
 
 
@@ -143,7 +150,7 @@ class ScalingStrategy(Enum):
     # DELAYED_SCALING = auto() # TODO: undetermined
 
 
-class ScalingRecipe(NamedTuple):
+class _ScalingRecipeFields(NamedTuple):
     """
     Supported MXFP8/MXFP4 scaling recipe.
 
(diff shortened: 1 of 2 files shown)


HuangWei-95:
## Review: LGTM, one test nit

Verified this is necessary and minimal. No conflict with `main` (the branch is one commit behind and touches nothing else).

I checked the fix against `pytorch` `release/2.11`: the `isinstance(cls, OpaqueBaseMeta)` check inside `register_opaque_type` in `torch/_library/opaque_object.py` is unconditional, so without this patch `import primus_turbo.pytorch` raises `TypeError` outright on torch 2.11. Adding only the metaclass without inheriting `OpaqueBase` is exactly what that error message prescribes, and it matters here because `ScalingRecipe` is a `NamedTuple` and could not take a second base class anyway.

The `try/except ImportError` is the right shape for forward compatibility too: pytorch `main` has since renamed this API to `CustomClassBase`, and when that lands the import will fail and `_OpaqueMeta` correctly falls back to `type`.

One thing to fix before merge:

```python
def test_torch_version_that_needs_the_metaclass_is_the_one_we_have():
    assert torch.__version__
```

This is vacuously true — `torch.__version__` is always a non-empty string, so the test passes on every torch version while its name promises a version check. Either assert the actual condition (e.g. that `OpaqueBaseMeta` is importable iff the metaclass is required) or drop the test; as written it will not catch a regression.

RuibinCheung (approved):
LGTM

PR #502, opened 2026-09-10
fix(flydsl): make the dense-TN split-K id decode exact

HuangWei-95:
## What breaks

The whole-loop dense-TN 4-wave FP8 kernel (`_dense_tn_wave4_tile`) splits the grid's last
partial round into a split-K window, and recovers `(slice, tile)` from a dispatch id by
dividing by the window width:

```python
win_off = d - lo
sid = _dense_tn_slice_div(win_off, nwin)       # nwin = tiles % ncu
tile_in_window = win_off - sid * nwin
```

`_dense_tn_slice_div(x, d)` was a fixed-point reciprocal pinned to a 16-bit shift:
`(x * ceil(2**16 / d)) >> 16`. With `mul = ceil(2**p / d)` and `r = mul * d - 2**p`, that form
equals `x // d` only while `xmax * r < 2**p`, and nothing checked the bound. The K-index divide
next to it has a guard (`_TN4_RCP_MAX`); this one did not.

So for a window of 192 tiles split 4 ways: `mul = 342`, `r = 342*192 - 65536 = 128`, exact only
below `x = 512`. Dispatch ids 575 and 767 read `sid = 3` instead of 2, giving
`tile_in_window = -1`. Tile 191 of slices 2 and 3 is never computed, two workgroups scribble a
degenerate tile, and the reduce pass then folds a workspace band nothing wrote. The scratch is
`torch.empty`, so the wgrad comes out NaN or finite garbage depending on what was in the buffer.

This is deterministic arithmetic, not a split-K race: which ids misdecode is a pure function of
`(nwin, s)`. It looks non-deterministic only because the value read from the unwritten band is
whatever the allocator last left there.

## The fix

Hand the dividend bound to the helper and let it pick the narrowest shift that is exact over
that range, asserting if the `i32` multiply would overflow. All four call sites now pass a
bound. The emitted code is the same multiply-and-shift pair with different immediates.

`_TN4_RCP_MAX` and its planner guard stay as they were, so the accepted split configurations do
not change; only their comments are reworded, since exactness is now the helper's job and that
bound is just keeping the multiply inside `i32`.

## Evidence

MI355X (gfx950, 256 CU), `torch 2.12.0+rocm7.15.0a20260727`, K=4096, per-tensor FP8 E4M3, TN
(`trans_a=True, trans_b=False`). 20 launches into a persistent output buffer, caches cleared
between shapes. `distinct` is how many bitwise-different results those 20 launches produced.

| C shape | route | (nwin, s) | reciprocal exact? | main | this PR |
|---|---|---|---|---|---|
| 2880 x 4096 | reduce | 192, 4 | no | **NaN**, distinct 20/20 | 52.49 dB, distinct 1 |
| 3072 x 4096 | reduce | 192, 4 | no | **NaN**, distinct 20/20 | 52.49 dB, distinct 1 |
| 4096 x 2880 | reduce | 165, 3 | no | **NaN**, distinct 20/20 | 52.44 dB, distinct 1 |
| 2560 x 2880 | handoff | 120, 2 | yes | 53.83 dB, distinct 1 | 53.83 dB, distinct 1 |
| 2880 x 2560 | handoff | 120, 2 | yes | 53.84 dB, distinct 1 | 53.84 dB, distinct 1 |
| 4096 x 4096 | no split | - | - | 55.59 dB, distinct 1 | 55.59 dB, distinct 1 |
| 2880 x 5120 | no split | - | - | 55.59 dB, distinct 1 | 55.59 dB, distinct 1 |

Failure tracks the arithmetic exactly: every shape whose window has an inexact reciprocal fails,
every shape whose window is exact passes, and the two shapes with no split-K window are
untouched. Forcing a handoff shape through the reduce pass instead (`_tn4_handoff -> False`)
also stays clean at 52.21 dB, so the in-kernel handoff route has no separate defect — both
routes are correct once the decode is.

### Cost

The change only moves two immediates, so this compares the patched tree against itself with
`_exact_recip` pinned back to a 16-bit shift: one process, one allocation set, blocks
interleaved `A B B A A B B A`, 300 iterations each timed with CUDA events, min over blocks.
The raced geometry winner was identical on both sides for every shape.

| C shape | exact shift | 16-bit shift | delta |
|---|---|---|---|
| 2880 x 4096 | 76.81 us | 79.26 us | -3.09% |
| 3072 x 4096 | 80.30 us | 80.48 us | -0.22% |
| 4096 x 2880 | 74.70 us | 74.89 us | -0.25% |
| 2560 x 2880 | 41.48 us | 41.80 us | -0.55% |
| 2880 x 2560 | 39.58 us | 39.63 us | -0.13% |

Four of five sit inside +/-0.6%. The -3.09% is one fast first block on the exact side (its other
three blocks are 79.4-79.9 us, level with the control), so read it as noise rather than a gain.
No measurable cost, as expected for a constant swap.

### Accumulating epilogue

`beta=1.0` never reaches this code: the wrapper gates the 4-wave path on
`_tn_wave4_supported(...) and not beta_is_one`, because a split-K window would fold an
accumulate in twice, so `beta=1` takes the 8-wave epilogue, which has no window. Measured at
55.59 dB and bitwise stable over 6 launches on all seven shapes above, on both trees, including
the three that are NaN at `beta=0`. Fused wgrad accumulation
(`fuse_bgrad_accum_pattern="megatron"`, i.e. Megatron `--gradient-accumulation-fusion`) is
therefore unaffected either way; it is the unfused wgrad, which writes a fresh tensor at
`beta=0`, that hits this.

## Exposure

Across a 9-model shape survey (dense projections, micro-batch 1 and 4), 72 wgrad GEMMs land on
the whole-loop TN path and 2 decode a wrong `(tile, slice)`: gpt-oss-20B `attn_out` at
micro-batch 1 and 4, which is C[2880, 4096], 192 tiles split 4 ways. A static model of the
planner picked the same winning geometry as the hardware on all four measured split shapes,
including the RECT winner at 4096 x 2880.

Two further scoping notes. The FP8 linear wgrad passes `default_backend=HIPBLASLT`, so FlyDSL is
reached only when it is pinned or the auto-tuner selects it. And the grouped-GEMM FlyDSL kernel
does not use this helper.

## Tests

`tests/pytorch/ops/test_flydsl_tn_split_index.py`, host-side, no GPU needed. It checks the
constants the kernel emits rather than re-deriving them:

- `_exact_recip` rounds down for every dividend up to the bound it was given, for `d` in
  1..1024, and its product stays inside `i32`.
- the split decode is a bijection for every `(nwin, s)` with `nwin` in 1..512 and `s` in
  `_TN4_SPLIT_S`.
- every window `_dense_tn_split` can emit, swept over tile counts up to `8 * ncu` for both macro
  tiles at 256 and 304 CU, decodes without collision.
- a regression lock on `nwin=192, s=4`, asserting a 16-bit reciprocal misdecodes exactly ids 575
  and 767.

9 passed. The last test's pre-fix assertion is what documents the old behaviour in-tree. For
scale: sweeping both macro tiles at 256 and 304 CU over K-iteration counts 16 to 128, the planner
can emit 387 distinct `(nwin, s)` windows, and a 16-bit reciprocal misdecodes 58 of them (15%) --
including `(165, 3)` and `(176, 4)`, which are the windows behind two of the failures measured
above. Over the test's full grid (`nwin` 1..512), 804 of 1536 pairs break.

The numbers in this PR come from driving `gemm_fp8_tensorwise_flydsl_kernel` directly, since the
prebuilt ROCm image I measured on has no matching `primus_turbo.pytorch._C`; CI covers
`tests/pytorch/ops/test_gemm_fp8.py` on a real build.


HuangWei-95:
### Reproducers

The arithmetic, no GPU and no imports -- this is the whole bug:

```python
nwin, s = 192, 4                        # C[2880, 4096] over the 256x256 tile, split 4 ways
mul = -(-(1 << 16) // nwin)             # 342, what _dense_tn_slice_div used
print([x for x in range(nwin * s) if (x * mul) >> 16 != x // nwin])
# [575, 767]   -> sid reads 3 instead of 2, so tile_in_window = -1
```

On hardware, against `main`. Needs only the FlyDSL tree on `PYTHONPATH`:

```python
import torch
from primus_turbo.flydsl.gemm.gemm_fp8_kernel import gemm_fp8_tensorwise_flydsl_kernel as tn

M, N, K = 2880, 4096, 4096
torch.manual_seed(1234)
s = torch.ones(1, dtype=torch.float32, device="cuda")
a = torch.randn(K, M, dtype=torch.bfloat16, device="cuda").clamp_(-4, 4).to(torch.float8_e4m3fn)
b = torch.randn(K, N, dtype=torch.bfloat16, device="cuda").clamp_(-4, 4).to(torch.float8_e4m3fn)
ref = (a.double().t() @ b.double()).float()
c = torch.empty((M, N), dtype=torch.bfloat16, device="cuda")

seen = []
for i in range(20):
    tn(a, s, b, s, trans_a=True, trans_b=False, out=c)          # TN, the wgrad layout
    torch.cuda.synchronize()
    snr = 10 * torch.log10(ref.pow(2).mean() / (ref - c.float()).pow(2).mean())
    if not any(torch.equal(c, x) for x in seen):
        seen.append(c.clone())
    print(f"launch {i:2d}  SNR {snr.item():7.2f} dB  distinct so far {len(seen)}")
```

`main`: `SNR nan` on every launch and 20 distinct results, because each launch folds whatever
the allocator left in the unwritten workspace band. This PR: `52.49 dB`, one result.

Note the persistent `out=` buffer. Calling without it allocates a fresh output each time, which
can mask the whole thing behind allocator reuse -- that is what made this look like a flaky
tolerance problem rather than a decode bug.

Two things the harness has to do or the picture is misleading. Clear
`_TN4_PLAN_CACHE / _TN_WAVE4_CACHE / _COMPILED_DENSE_CACHE / _TN4_WS_CACHE / _TN4_FLAG_CACHE`
between shapes, since the split-K scratch is cached and a broken shape's leftovers otherwise
follow the next one. And read the geometry the race actually settled on out of
`_TN_WAVE4_CACHE`, because the two macro tiles give different `nwin` for the same C shape.

### Raw measurements

<details><summary>Accuracy and accumulate, main vs this PR (20 launches, persistent buffer)</summary>

```
shape             route     nwin,s   main                        this PR
2880x4096x4096    reduce    192,4    nan dB, distinct 20/20      52.49 dB, distinct 1
3072x4096x4096    reduce    192,4    nan dB, distinct 20/20      52.49 dB, distinct 1
4096x2880x4096    reduce    165,3    nan dB, distinct 20/20      52.44 dB, distinct 1
2560x2880x4096    handoff   120,2    53.83 dB, distinct 1        53.83 dB, distinct 1
2880x2560x4096    handoff   120,2    53.84 dB, distinct 1        53.84 dB, distinct 1
4096x4096x4096    no split   -       55.59 dB, distinct 1        55.59 dB, distinct 1
2880x5120x4096    no split   -       55.59 dB, distinct 1        55.59 dB, distinct 1

beta=1.0 (fused wgrad accumulation), 6 launches, every shape, both trees:
  55.59 dB, distinct 1
```

The winning geometry per shape, as raced on the hardware: `2880x4096` and `3072x4096` take
256x256; `4096x2880` takes 384x192, which is why its window is `(165, 3)` and not `(192, 4)`.

</details>

<details><summary>Cost, patched tree against itself with the shift pinned back to 16 bits</summary>

Blocks interleaved `A B B A A B B A`, 300 iterations per block timed with CUDA events, one
process, same allocations. `us` lists the four blocks per side; the delta compares the minima.

```
shape             exact shift (us)              16-bit shift (us)             delta
2880x4096x4096    76.81 79.92 79.42 79.79       79.26 79.68 81.78 79.40       -3.09%
3072x4096x4096    80.55 80.82 80.38 80.30       80.53 81.35 83.30 80.48       -0.22%
4096x2880x4096    76.66 76.33 74.70 75.19       74.89 75.27 74.97 74.94       -0.25%
2560x2880x4096    41.78 42.24 41.55 41.48       41.80 42.12 42.07 41.71       -0.55%
2880x2560x4096    39.77 40.01 40.41 39.58       40.19 39.91 39.63 39.77       -0.13%
```

Same raced winner on both sides for every shape. An earlier pass with only two blocks per side
read `+4.95%` on `2880x2560`; that block's own spread was 8%, and it lands at `-0.13%` here.

</details>

<details><summary>How the shift is chosen, and what it costs in instructions</summary>

```
window (nwin, s)   pre-fix (shift, mul)   this PR (shift, mul)   ids misdecoded pre-fix
(192, 4)           16, 342                17, 683                2   [575, 767]
(176, 4)           16, 373                17, 745                1   [703]
(165, 3)           16, 398                17, 795                1   [494]
(120, 2)           16, 547                14, 137                0
(105, 2)           16, 625                15, 313                0
```

`_exact_recip` returns the *narrowest* exact shift, so the constants move for
already-correct windows too -- `(120, 2)` and `(105, 2)` end up with a smaller multiply than
before, not a larger one. Either way it is one `v_mul_lo_u32` and one shift, same as before, so
the only thing that changes is an immediate. That the two handoff shapes still land on exactly
53.83 and 53.84 dB with one distinct result each is the check that renumbering them changed
nothing.

</details>

RuibinCheung (approved):
LGTM

PR #505, opened 2026-09-14
perf(flydsl): speed up the MXFP8 dense GEMM and dual-cast quant on gfx950

kyle-256:
# Description

MXFP8 dense GEMM only has an NT kernel, so dgrad (NN) and wgrad (TN) are expressed as NT
against a materialised transpose, while per-tensor FP8 has native NN/TN kernels. Measured
over a full training step with MXFP8's dual-cast quantisation charged to MXFP8, MXFP8 was
slower than per-tensor FP8 on three of the five production linears.

This PR closes that gap and passes it: MXFP8 is now faster than per-tensor FP8 on every
one of the five shapes, geomean +10.9%. Kernel-only changes -- no change to numerics,
formats, block sizes or shapes.

## Type of change

Performance optimisation (non-breaking). Closest box below:

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

`primus_turbo/flydsl/gemm/gemm_mxfp8_kernel.py`

- Pack four K-groups of E8M0 scales per preshuffled dword and select the byte with the
  scaled-MFMA `op_sel`. Unpacked, a dword carried the single byte one K-iteration needs,
  so the preshuffle wrote 4x the plane it read and the GEMM streamed 4x the scale bytes.
- Dense-local widened scale preshuffle. The shared repack in `gemm_helper.py` also serves
  the grouped and mega callers and cannot widen its staging; this dense-only mirror stages
  `pack` dwords per lane and byte-permutes the packed dword.
- Merge a lane's two n-fragments into one granule-filling C store, either by
  `v_permlane16_swap_b32` or by pairing B columns so the fragments are already adjacent.
- Tail-round split-K. A dispatch runs `ceil(tiles/NCU)` serialised rounds and the last is
  only `tiles%NCU` wide; splitting K for the tail tiles alone refills that round at a
  partial-sum cost proportional to the tail, not to the whole output.
- Lap the C-store epilogue by quadrant into the last K-iteration so it rides an MFMA
  shadow, and keep the SIMD-mate wave pair on different `wave_m` so they share B, not A.
- C-store cache policy added as an autotuned axis (its sign is shape-dependent).

`primus_turbo/flydsl/quantization/mxfp8_quant_flydsl.py`

- Split the store cache policy into separate scale and data axes. The E8M0 stores are
  byte-granular and rely on L2 merging the write masks, so a non-temporal hint there turns
  every byte into a partial-line round trip.
- Long-run walk for casts scattered on both padded dims: `krun` lengthens the input read
  and the row write, an XCD-owned pid band lengthens the column write.
- vec8 load path when `K % 8 == 0`, and a 2D-capable warp-reduce layout that produces both
  the per-32 and the shared 32x32 scale via `ds_bpermute`.

## Performance

Measured on MI355X (gfx950), one GPU, whole training step (forward + real `.backward()`)
per linear, with MXFP8's dual-cast quantisation included. TFLOP counts all three GEMMs of
the step (fwd + dgrad + wgrad).

| linear | M, N, K | TFLOP | share | before | after | gain |
|---|---|---|---|---|---|---|
| QKV | 8192, 9216, 3072 | 1.392 | 6.8% | 1836 TF/s | **2037 TF/s** | +11.0% |
| MLP-up | 8192, 12288, 3072 | 1.855 | 11.9% | 1837 TF/s | **2111 TF/s** | +14.9% |
| MLP-down | 8192, 3072, 12288 | 1.855 | 12.0% | 1844 TF/s | **2071 TF/s** | +12.3% |
| linear1 | 16384, 21504, 3072 | 6.494 | 38.6% | 2192 TF/s | **2379 TF/s** | +8.5% |
| linear2 | 16384, 3072, 15360 | 4.639 | 30.8% | 2175 TF/s | **2361 TF/s** | +8.6% |

Accuracy is unchanged: every scoring run held `snr_min` at 28.3 dB against a 27.27 dB
floor, and the preshuffle and quantiser changes were verified bit-identical against the
previous implementation.

# Checklist:

- [ ] I have made corresponding changes to the documentation (no documentation affected)


RuibinCheung (approved):
LGTM

PR #510, opened 2026-09-15
fix(jax): stop Unknown architecture failures when HSA is exhausted

HuangWei-95:
## Summary
- JAX `quantize_fp8` abstract eval was calling `warp_size()` → `hipGetDeviceProperties`. That is a host-side shape query and must not require a healthy GPU. On the long-lived `turbo_jax_github_runner_v2605` container, leftover `pytest -n 8` workers hold `/dev/kfd`; with the image default `GPU_MAX_HW_QUEUES=2`, the next `rocminfo`/HIP init returns `HSA_STATUS_ERROR_OUT_OF_RESOURCES`, JAX falls back to CPU, and col-major FP8 tests fail as `Unknown architecture`.
- Confirmed on `smc300x-ccs-aus-a16-10`: host `rocminfo` can still work while the runner container's `rocminfo` fails, 8 xdist workers hold kfd, host VRAM stays 0%, and a fresh `import jax` reports `HIP_ERROR_NoDevice`. This is occupancy-dependent, not a random gfx942 mis-detect, and not caused by #501.
- Size col-reduce workspace without querying the GPU, stop caching a failed HIP probe as `UNKNOWN`, isolate xdist workers on HIP/ROCR/CUDA, force `JAX_PLATFORMS=rocm`, reclaim leftover HSA clients at job start, and fail CI fast if ROCm is unavailable.

## Test plan
- [ ] JAX CI `unittest-jax-gfx942`: `Check ROCm GPU` sees `gfx942` after reclaim; `Check JAX ROCm backend` prints a non-CPU backend
- [ ] `tests/jax/lax/test_quantization.py` col-major cases (`axis != -1`, including `[1,111,4096]`) no longer raise `Unknown architecture`
- [ ] Re-run on the JAX runner after a cancelled/timed-out `pytest -n 8` job and confirm leftover kfd clients are reaped before tests start

Made with [Cursor](https://cursor.com)


csrc/include/primus_turbo/reduce.h
```diff
@@ -28,8 +28,12 @@ int64_t get_reduce_row_workspace_sizes(const int64_t &outer_len, const int64_t &
...
 template <typename ComputeType>
 int64_t get_reduce_col_workspace_sizes(const int64_t batch, const int64_t m, const int64_t n) {
-    const int BLOCK    = 256;
-    const int NUM_WARP = BLOCK / warp_size();
+    const int BLOCK = 256;
+    // Host-side size query must not call warp_size()/HIP. JAX abstract eval
+    // runs before a device is bound; a failed HIP query used to throw
+    // "Unknown architecture". WAVE=64 is the CDNA default; on wave32 this
+    // over-allocates (fewer warps → larger buffer), which is safe.
+    const int NUM_WARP = BLOCK / 64;
```
RuibinCheung: For gfx1250, the warp size is 32. Should use warp_size() to get runtime warp size.

RuibinCheung (changes requested):

csrc/include/primus_turbo/reduce.h
HuangWei-95: Restored `NUM_WARP = BLOCK / warp_size()` so gfx1250 (wave32) sizes the col-reduce workspace from the runtime warp size. HIP-unavailable cases stay in `warp_size()` (no throw / no caching UNKNOWN).

PR #513, opened 2026-09-15
ci: clean stale submodule checkout before recursive clone

WangLingxun:
## Summary
- Self-hosted GPU runners reuse `_work`. A cancelled `git submodule update` can leave `3rdparty/composable_kernel` git metadata (including `index.lock`), and the next `actions/checkout` with `submodules: recursive` fails with exit 128.
- Wipe leftover `3rdparty/` and `.git/modules` after `Change owner` and before checkout on the torch/jax unit-test jobs, benchmark, and release sdist build.


RuibinCheung (approved):
LGTM

